PR #19, opened 2020-12-08
Dimension unit tests

DrDonk:
A set of unit tests that test most of the API.
Also 2 fixes for dimension proxy in plaid-utilities.


DrDonk:
These tests will never be able to run in Jenkins. Where to put them?

PR #49, opened 2021-02-09
disable_variables flag

adamsdarlingtower:
Allow variable application to be disabled for the parts of sql_expression data_explorer needs.

There will be another data_explorer PR once this is in - if we try to do them both at the same time, linting will fail


chucklay (approved):
LGTM

PR #55, opened 2021-03-09
run pylint and report results

gbates101:
Reference: [ch6860](https://app.clubhouse.io/plaidcloud/story/6860/plaid-utilities-plaid-rpc-no-longer-run-the-unit-tests)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -33,6 +33,7 @@ podTemplate(label: 'plaid-utilities',
         if (!params.skip_lint) {
           sh """
             lint --target-dir=$params.target_lint_dir --branch=$branch --full-lint=$params.full_lint
+            pytest
           """
 
           if (branch == 'master') {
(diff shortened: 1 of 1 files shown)


rad-pat:
Looks good, but the tests seem to require the UTF8 locale setting up so they're failing. The lint is also failing on this same as plaid and I didn't know where to look for the build image. It's something to do with how we're referencing the branch in the command sent to lint, however it does seem that it is working ok for workflow-runner. 
```
Linting project... (This could take several minutes)
fatal: ambiguous argument 'origin/master...run-unit-tests': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

gbates101:
> Looks good, but the tests seem to require the UTF8 locale setting up so they're failing. The lint is also failing on this same as plaid and I didn't know where to look for the build image. It's something to do with how we're referencing the branch in the command sent to lint, however it does seem that it is working ok for workflow-runner.
> 
> ```
> Linting project... (This could take several minutes)
> fatal: ambiguous argument 'origin/master...run-unit-tests': unknown revision or path not in the working tree.
> Use '--' to separate paths from revisions, like this:
> 'git <command> [<revision>...] -- [<file>...]'
> ```

This error is also taking place in workflow-runner builds, so I'm surprised that linting is working for workflow-runner at this point. Not sure what part of the git command syntax is leading to this error.

rad-pat:
I wonder if the diff should be something like `git diff --name-only --diff-filter=AM origin/master..origin/branch`
specifying `origin/` for the PR branch

PR #57, opened 2021-03-24
Use asyncio and orjson to convert JSON to CSV

rad-pat:
For use with plaid PR https://github.com/PlaidCloud/plaid/pull/3493


inviscid (approved):
LGTM

PR #58, opened 2021-03-25
Put this back to sync code, aiocsv is junk

rad-pat:
I guess not all aio libraries are that great, aiocsv is not performant, nor does it read csvs in the same way that the standard lib does. moves the conversion back to synchronous code that can be called in an executor


gbates101 (approved):
LGTM. Does running in an executor avoid blocking, or would that require an asyncio thread or subprocess call?

rad-pat:
Y, running in a thread executor does avoid blocking for IO based tasks

PR #59, opened 2021-04-02
[WIP] Module to parse consolidated helm chart config

gbates101:
### What This Does

Parses a file of _common configuration options_ across our applications, mounted by a [consolidated helm chart](https://github.com/PlaidCloud/k8s/pull/164) to each pod in each app deployment. This allows a marketplace deployment of our application stack via helm to dictate the redis/rabbitmq/postgres/etc instances and their credentials to be used by our applications in that deployment. These settings include:

- Feature gates, which may be used cross-application.
- Redis/RMQ/Database credentials and connection info.
- Service URLs, as they are dictated by the helm release name and are subject to change from deployment to deployment.
- Global environment settings, such as the designation (dev, prod, qa, beta), and the hostname being used.
- Possibly other things that I have failed to realize are important to include.

### What This Doesn't Do

Not intended to provide configuration options for each specific applications, such as flask/starlette configurations, credentials to be managed and mounted as separate files (RSA keys, SSL certs, etc), and anything else that would not be versioned in a public repo or distributed publicly. 

### Notes on Implementation

Pretty straightforward implementation that leverages classes that extend `NamedTuple` (so your text editors can leverage type hinting for config objects), and loads the config file sections in as kv pairs and returns the appropriate instance to the caller.

### Questions

1. Should the configuration file being loaded be validated by some public schema?
2. Are there any cross-app configuration settings that I'm missing or would be useful?
3. Is this repo an appropriate publication site for this module?
4. General feedback? Will be made available to all of our apps, so likely (hopefully) won't change much post-merge.

Expanding on question 3, my thinking was to provide this module here so clients that self-host our application stack (citing [this Clubhouse task](https://app.clubhouse.io/plaidcloud/story/7307/enable-plaidcloud-deployment-as-google-marketplace-app)) who intend to write third-party apps could more comfortably integrate with our applications.  

[Here is a link](https://github.com/PlaidCloud/k8s/blob/814a7b87364acdddb8954c8f7da9168b3a9ac318/charts/plaidcloud/templates/configmap.yaml#L9-L53) to the configuration file as defined in the WIP helm chart found here: https://github.com/PlaidCloud/k8s/pull/164

### DIFF
--- a/plaidcloud/utilities/config.py
+++ b/plaidcloud/utilities/config.py
@@ -59,6 +59,7 @@ class RedisURLConfig(NamedTuple):
     identity_cache: str = "redis://redis-master/8"
     ipython_registry: str = "redis://redis-master/3"
     oauth_cache: str = "redis://redis-master/9"
+    redis_scheduler: str = "redis://redis-master/13"
     scopes_cache: str = "redis://redis-master/12"
     session: str = "redis://redis-master/0"
     transform_container_registry: str = "redis://redis-master/5"
@@ -119,7 +120,7 @@ class PlaidConfig:
 
     def redis_client(self) -> RedisConfig:
         """Settings for Redis client connections."""
-        redis_config = = self.cfg.get('redisClient', {})
+        redis_config = self.cfg.get('redisClient', {})
         return RedisConfig(**redis_config)
 
     @property
--- a/plaidcloud/utilities/config.py
+++ b/plaidcloud/utilities/config.py
@@ -1,0 +1,139 @@
+3#!/usr/bin/env python
+# coding=utf-8
+"""Loads the configuration file used by workflow monitor.
+"""
+import os
+import yaml
+from typing import NamedTuple
+
+CONFIG_PATH = os.env.get('PLAID_CONFIG_PATH', '/etc/plaid/config.yaml')
+
+
+class DatabaseConfig(NamedTuple):
+    hostname: str
+
+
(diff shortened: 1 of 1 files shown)


adamsdarlingtower:
Just started looking at this,  and I suspect I'll have more comments.

If this is supposed to be available to all our apps, this repo is the wrong place for it. Part of the intent of this repo is that it should be only installed in apps that need it (currently only plaid and workflow-runner). plaid-rpc is the one that's installed on all apps

adamsdarlingtower (changes requested):
I don't know of other cross-app configuration settings, but they might exist. I do think we should have a plan for future changes, such that things won't break. That's a little harder with NamedTuple than with dict, since dicts don't have a set order or set length.

With regards to a public schema - yeah, it would probably be a good idea, but I don't think it's absolutely necessary

gbates101:
@adamsdarlingtower the contents of this PR have been moved to the [plaidcloud-config](https://github.com/PlaidCloud/plaidcloud-config/tree/master/python) repo, structured in a way to allow for parsers written in other languages (javascript?) as the need arises. Public schema has not been implemented yet, but I do agree it would be a good move. Third parties _might_ edit the config file when self-hosting, so having 'guard rails' up would be helpful.

Regarding the other issues, they have been addressed in the other project. Since this code has been moved entirely, I'm closing this PR.

PR #63, opened 2021-04-08
Give further detail when apply_variables does not work

rad-pat:
* What was the string we were trying to replace variables in

This is currently raising Exception in IO, but no idea why. The same step runs fine for me locally.


chucklay (approved):
LGTM

PR #72, opened 2021-05-17
Ch7410 - XLRD/OpenPYXL

rad-pat:
So, it seems XLRD3 is slightly faster than openpyxl for some things, specifically iterating a whole document to convert to CSV. We already use openpyxl for some other things such as to get a list of worksheets quickly so it's not going away. Here, I just add the equivalent convert_to_csv method in case we may want to use it in future.


inviscid (approved):
LGTM

PR #73, opened 2021-05-17
add binary dtype to handle inbound binary columns

inviscid:


rad-pat (approved):

rad-pat:
Does this need to be amended in here to handle binary type? https://github.com/PlaidCloud/plaid-rpc/blob/30975305ac333f2939c14130632ca2d6636dfeca/plaidcloud/rpc/type_conversion.py#L22

inviscid:
Good catch.  PR for that repo now open.

PR #75, opened 2021-05-19
handle additional sql dtypes that are common

inviscid:


plaidcloud/utilities/frame_manager.py
```diff
@@ -115,12 +116,25 @@ def sql_from_dtype(dtype):
...
+        'inet': 'text',
+        'macaddr': 'text',
     }
 
-    if str(dtype).lower().startswith('numeric'):
+    dtype = str(dtype).lower()
+    if dtype.startswith('num'):
         dtype = 'numeric'
+    elif 'char' in dtype:
+        dtype = 'text'
```
rad-pat, line 135: As we have some doc tests, we should probably should add a couple of test cases in the docstring for the cases where it begins num, or contains char
```
>>> sql_from_dtype('number')
'numeric'
>>> sql_from_dtype('varchar(123)')
'text'
```

rad-pat (changes requested):
Update Doc Test

rad-pat (approved):
Lovely

PR #85, opened 2021-07-22
alt method of applying rules: multiple rule matches = multiple results

rea725:


rad-pat:
Now I can see what you're after here, I can see that the `include_once` flag set False *should* do the job that you're after, but doesn't and hence you're having to add this virtual duplicate of the matching code. I'm confident that I can fix the `include_once` to work as it should in the case that it is set False.

rad-pat (approved):

PR #91, opened 2021-09-22
Ch7980

adamsdarlingtower:
Relative to the previous ch7980 PR there are a few changes. In particular, an attempt to dynamically handle multiple sources for arguments in get_select_query has been replaced with minimal boilerplate code. Besides that, the order in which cast_fn and agg_fn are applied to columns, etc. are reversed. This should fix the issues we were having with table explrorer.

This needs to be merged and tagged before PRs for other services go in, so they can point at the right tag. I'm not sure if this should actually go into master - should we make sure that whatever is currently in beta and prod are pointing at a tag too, before that happens?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist
 __pycache__
 dim_current/
 .plaid/
+flycheck*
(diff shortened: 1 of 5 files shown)


rad-pat (approved):
Looks good. Yeah, probably plaid, workflow-runner & data-explorer need to be updated in Beta with versions that point to v1.0.0 and then this can be merged to master

PR #93, opened 2021-09-27
Fix the update transform issue

adamsdarlingtower:
Distinguish between returning "This column should be NULL" and "This column shouldn't be included"


inviscid (approved):
LGTM

PR #97, opened 2021-10-07
Ch8385

adamsdarlingtower:
Adds unit tests to sql expression. The majority of them pass for both v1.0.0 and the code included in this PR.

Those that don't:
* All tests of import_data_query error in v1.0.0, because import_data_query generates a uuid for the name of its temp table. I modified it to allow the caller to pass in a temp table name, for testing purposes, so that the output is predictable.
* All tests of get_update_value error in v1.0.0, because the function doesn't exist. I separated it out in order to both make the code for get_update_query cleaner, and also to make testing simpler and more thorough.
* In v1.0.0, TestGetSelectQuery.test_put_columns_without_order_at_end_of_sort errors, because columns with a sort param but not an order param cause a KeyError. Columns with a sort param but no ascending param are just ignored for order_by. I've added what I think is a reasonable behavior, if columns without a sort order are passed in - they get put at the end, after any columns that do have a sort order. I'm not actually sure it's possible to send in a column with a sort param but no order param in the UI.
* in v1.0.0, TestGetColumnTable.test_table_dot_column errors, despite code to handle that specific case of 'table.column', because the code to handle 'tableN.column' catches it and then gets confused when there's no N. In the current, I've reversed the order of those handlers, so we can get what seems to be the intended behavior.

Besides that, this PR:
* removes the unused function apply_rules
* adds explanatory comments in several places
* fixes places where v1.0.2's behavior did not match v1.0.0
* adds a default value for the metadata param to get_table_rep_using_id, in order to match the api of get_table_rep, which has that default value.
* fixes the default value for the variables param in apply_output_filter, so using that default value no longer causes an error
* makes import_data_query non-destructive
* adds some refactoring TODO comments

### DIFF
--- a/plaidcloud/utilities/sql_expression.py
+++ b/plaidcloud/utilities/sql_expression.py
@@ -219,6 +219,7 @@ def get_from_clause(
             source_without_table = source  # a couple extra checks, in an error scenario, but flatter code
 
         if target_column_config.get('agg') == 'count_null':
+            #TODO: should this also check the "aggregate" param?
             col = None
         elif source in table.columns:
             col = table.columns[source]
@@ -550,11 +551,11 @@ def get_select_query(
         select_query = sqlalchemy.select(*column_select)
 
     # Build WHERE section of our select query
-    wheres = [w for w in wheres if w] if wheres else []
-    if wheres:
-        combined_wheres = get_combined_wheres(
-            wheres, tables, variables, disable_variables, table_numbering_start=table_numbering_start
-        )
+    wheres = wheres or []
+    combined_wheres = get_combined_wheres(
+        wheres, tables, variables, disable_variables, table_numbering_start=table_numbering_start
+    )
+    if combined_wheres:
         select_query = select_query.where(*combined_wheres)
 
     # Find any columns for sorting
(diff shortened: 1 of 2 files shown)


plaidcloud/utilities/sql_expression.py
```diff
@@ -258,6 +260,7 @@ def get_agg_fn(agg_str):
     if hasattr(sqlalchemy.func, agg_str):
         return getattr(sqlalchemy.func, agg_str)
 
+    # ADT2021 - I believe this line is unreachable, because hasattr(sqlalchemy.func, _) should always be True
```
rad-pat, line 263: Yep, totally pointless having the check and the `return ident`. It should probably just read `return getattr(sqlalchemy.func, add_str)` and omit the always True check

rad-pat (changes requested):
I think removing the inaccessible line of code makes sense.
Do you know if the changes will fix https://app.shortcut.com/plaidcloud/story/8430/bug-error-in-io-psycopg2-errors-syntaxerror-non-integer-constant-in-order-by
I guess we may have to merge to find out.

adamsdarlingtower:
> I think removing the inaccessible line of code makes sense. Do you know if the changes will fix https://app.shortcut.com/plaidcloud/story/8430/bug-error-in-io-psycopg2-errors-syntaxerror-non-integer-constant-in-order-by I guess we may have to merge to find out.

I'm not certain if it will - I suspect the change back to 1.0.0's constant casting behavior may fix this. I think merging to find out is the right thing, and if it doesn't, that bug is what I'm working on next.

rad-pat (approved):

PR #114, opened 2021-12-02
Ch8649

adamsdarlingtower:
A lot of changes remove python 2 support, but the most important commits are:

```Moves IPython import inside the only function that uses it```
```Moves openpyxl import inside the only (unused) function that uses it```

These move heavy imports inside the functions that use them, so that they're only imported when they're needed. There's a lot of benefit to that in my testing methodology, though less in a production system where pods live longer. There could be greater benefit to importing everything when a pod spins up before we start listening for requests.

### DIFF
--- a/plaidcloud/utilities/data_helpers.py
+++ b/plaidcloud/utilities/data_helpers.py
@@ -8,11 +8,13 @@ import re
 import math
 import platform
 import locale
+import functools
+
+import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


chucklay (approved):

rad-pat (approved):
Looks ok, wonder why tests are failing

rad-pat (approved):

PR #117, opened 2021-12-22
Dim helpers

rea725:
Helper methods so we don't have to do this sort of thing within UDFs

### DIFF
--- a/plaidcloud/utilities/query.py
+++ b/plaidcloud/utilities/query.py
@@ -13,11 +13,10 @@ from sqlalchemy.dialects.postgresql.base import PGDialect
 from sqlalchemy_hana.dialect import HANABaseDialect
 from sqlalchemy_greenplum.dialect import GreenplumDialect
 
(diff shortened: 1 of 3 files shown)


plaidcloud/utilities/query.py
```diff
@@ -428,6 +425,30 @@ def bulk_insert_dataframe(self, table_object, df, append=False):
...
+                    sep='\t',
+                    encoding='UTF-8',
+                    quoting=csv.QUOTE_MINIMAL,
+                    escapechar='"'
+                )
+            )
+
+            # logger.info('----- CSV DATA DIRECTLY FROM PANDAS---- {}'.format(cs.getvalue()))
+            return cs.getvalue()
+
```
rad-pat, line 451: I think this was deleted in a previous PR, somehow it wants to put it back. I guess that was a bad merge or something.

rad-pat: Looks like it was the first commit in this PR that removed the use of six package. This code can likely be removed safely.

rad-pat (changes requested):
Just need to remove some code hanging around that I don't think is used.

PR #122, opened 2022-01-18
minor improvement.  No longer need if statement around every use.

rea725:


plaidcloud/utilities/udf.py
```diff
@@ -1,5 +1,7 @@
 from pathlib import Path
 from plaidcloud.rpc.config import find_workspace_root
+import getpass
+import os
```
inviscid, line 4: os import seems to be unused

inviscid (changes requested):

rad-pat (commented):
I was slightly confused by the title which says no longer require if, but it's adding an `if`. 
It took me a minute, but I get it now. Looks ok but for @inviscid comment.

PR #133, opened 2022-10-13
Ch10590

adamsdarlingtower:
Adds casts to functions used as cleaners in transforms


plaidcloud/utilities/sqlalchemy_functions.py
```diff
@@ -359,53 +397,144 @@ def compile_safe_to_date(element, compiler, **kw):
...
 class sql_safe_divide(GenericFunction):
     name = 'safe_divide'
 
-
 @compiles(sql_safe_divide)
 def compile_safe_divide(element, compiler, **kw):
     """Divides numerator by denominator, returning NULL if the denominator is 0.
     """
     numerator, denominator, divide_by_zero_value = list(element.clauses)
+    numerator = func.cast(number, sqlalchemy.Numeric)
```
rad-pat, line 536: Where does number com from?

plaidcloud/utilities/sqlalchemy_functions.py
```diff
@@ -359,53 +397,144 @@ def compile_safe_to_date(element, compiler, **kw):
...
+    denominator = func.cast(digits, sqlalchemy.Integer)
```
rad-pat, line 537: Where does digits come from?

rad-pat (changes requested):
I think the only problems are the undefined variables?

plaidcloud/utilities/sqlalchemy_functions.py
adamsdarlingtower: ah, good catch, thank you

rad-pat (approved):

PR #151, opened 2023-02-17
allow JSON and UUID as real dtypes

inviscid:


rad-pat (approved):
The tests should now pass when updated with latest master

PR #152, opened 2023-02-21
Use parse_dates for timestamps

rad-pat:


rea725 (approved):
🙈 🙊 🙉   so, did it break due to a SQLAlchemy change?  Just curious.

rad-pat:
I don't know if it broke, or that the Last Modified field is a new field, or perhaps the field type is different to before?

PR #157, opened 2023-04-28
Sc 12381

rad-pat:
Use row_number() over (order by) to insert values in serial columns in correct order.
Until I get a response from someone on this, we have to avoid the use of serial on GP and insert the values ourselves as serial is not working as we would expect inserting ordered data.


rad-pat:
I need to fix the tests

inviscid (approved):

PR #163, opened 2023-09-19
Sc 5058

chucklay:


plaidcloud/utilities/frame_manager.py
```diff
@@ -2648,6 +2649,20 @@ def fixedwidth_to_csv(fixed_width_file_name, csv_file_name, colspecs):
     )
 
 
+def parquet_to_csv(parquet_file_name, csv_file_name, start_row=0):
+    pf = ParquetFile(parquet_file_name)
+    df = pf.to_pandas()
+    if start_row:
+        df.iloc[start_row:]
```
rad-pat, line 2656: I think this prob wants to be df = df.iloc[start_row:]

rad-pat (changes requested):
I think the iloc statement is not right.
Pandas will read/write parquet file too, is the library better, or does pandas use it anyway? https://pandas.pydata.org/docs/reference/api/pandas.read_parquet.html#pandas.read_parquet

rad-pat (approved):

PR #164, opened 2023-09-21
Add source_table_name processsing

rad-pat:
Uses the table alias (or the actual table name if an alias is not specified) to populate the source_table_name column


inviscid (approved):

inviscid:
i approved but there are test failures it appears

inviscid (approved):

PR #167, opened 2023-11-14
Sql zfill and normalize_whitespace

adamsdarlingtower:
Adds both zfill function and normalize_whitespace function. They've been tested by making sure they compile to sql that looks right.


rad-pat (commented):
Looks ok, can you add some tests for these functions please?

rad-pat (approved):

PR #183, opened 2024-05-20
Databend functions

adamsdarlingtower:


inviscid:
@adamsdarlingtower have a look at the linter complaints and see if any of them are a concern
@rad-pat - We need to maintain this on a separate branch for now, correct?  Do we need to have a `databend-master` or something so we can keep things clean until we are fully ready to make the jump?

rad-pat:
> @adamsdarlingtower have a look at the linter complaints and see if any of them are a concern @rad-pat - We need to maintain this on a separate branch for now, correct? Do we need to have a `databend-master` or something so we can keep things clean until we are fully ready to make the jump?

I don' think we need a different branch, these changes will only affect databend connections

inviscid:
@adamsdarlingtower can you review based on Pat’s comments

PR #185, opened 2024-06-07
Update plaidcloud-rpc to v1.4.0

adamsdarlingtower:


inviscid:
Looks like it is failing build because it can't find the plaid-rpc tag specified

PR #218, opened 2024-11-21
Add provision for grouping_sets and rollups

rad-pat:
This is very basic initial implementation, will rollup, cube or create grouping sets on the columns as per group by.


inviscid (dismissed):

rad-pat:
Updated to add a fix for Mike running UDFs locally. The `auth_token` may be a callable method, in which case we should call it.

chucklay (approved):
LGTM

PR #224, opened 2025-02-04
Sc 19172 - Check for same datatype in pandas column

rad-pat:
Reports a more helpful error message than the pyarrow error.


rea725 (approved):
Is this to fix KC?

rad-pat:
Well, not to "fix", but to help diagnose more easily.

PR #241, opened 2025-04-23
Allow disabling of casting in get_select_query

chucklay:


rad-pat:
I see this is for preventing truncation for Snowflake, don't we just need a default type descriptor compiler for snowflake dialect? I think we had to do similar for Databend. Removing the cast may be useful anyway also though.

rea725 (approved):

chucklay:
@rad-pat Good call, forgot that was an option.

PR #243, opened 2025-05-19
Changes for UDF Parameters

rad-pat:
Allows UDFs to access source and tables as `con.udf.source[0]` `con.udf.target[0]` `con.udf.variables['myvariable']`


rea725 (dismissed):

rad-pat:
Fixed licence check issue

rea725 (approved):
Curious why we need an exact version of setuptools but OK.

rad-pat:
License checker is just failing to read the licence for some reason, we're not pinned to any specific version.

PR #246, opened 2025-05-22
Allow access UDF params by name

rad-pat:


inviscid (approved):
Excellent.  Will get this into use quickly.  The list approach is also very useful because the source name may change for reusability.

rad-pat:
I forgot to check that the ids are actually called 'id' - will check that first 🫨

PR #265, opened 2025-09-17
Remove get_table_rep_using_id, use get_table_rep instead.

simozzy:


rad-pat (changes requested):
Might want to bump the version in pyproject.toml in this PR and then we can make a release when it is merged, otherwise the release will fail

simozzy:
> Might want to bump the version in pyproject.toml in this PR and then we can make a release when it is merged, otherwise the release will fail

Thanks.  done.

PR #288, opened 2026-03-26
Sr db compatability

inviscid:


inviscid:
@claude review

plaidcloud/utilities/starrocks_compat.py
```diff
@@ -0,0 +1,3375 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+# PostgreSQL → MySQL/StarRocks date-format token mapping.
+# Sorted by descending length at lookup time so that longer tokens
+# (e.g. ``HH24``) are matched before shorter ones (``HH``).
+_PG_TO_MYSQL_FORMAT = {
```
rad-pat: I think there's already a helper for this, maybe named PG to python

plaidcloud/utilities/starrocks_compat.py
```diff
@@ -0,0 +1,3375 @@
...
+from sqlalchemy.exc import CompileError
+from sqlalchemy.sql import functions
+from sqlalchemy.ext.compiler import compiles
+from sqlalchemy.types import (
+    BigInteger, Boolean, DateTime, Date, Integer, Numeric, String,
+)
+
+_log = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
```
rad-pat: I think this wants to work the other way around, so that when we import sqlalchemy_functions, this module also gets imported. If the functions we already made are not substantial enough then we should just move those implementations over here and use the Claude-generated versions.

rad-pat (changes requested):

plaidcloud/utilities/starrocks_compat.py
rad-pat: postgres_to_python_date_format, or vice-versa (in plaidcloud-rpc type_conversion)

inviscid:
@rad-pat see if this is more aligned with your thinking

rad-pat (approved):
Yes, looks better. I see there's actually some difference between python and mysql date formatting, who knew?!

rad-pat:
Oh wait, will need to also bump the version in pyproject.toml so it can be released and used.

inviscid:
Closing as **superseded** — same situation as cp-rest#95.

**Why:** while this branch sat (~2 months, master moved 8 commits), the team built and tested its own StarRocks support directly in `plaidcloud/utilities/sqlalchemy_functions.py`: `e9d1344 Fix some functions for starrocks`, `a1cafc1 Ensure numerics … for Starrocks`, `53ee549 Fix boolean on Starrocks import`, `bed2af2 Fix round …`, `ac62dad fix(import_col): wrap col in to_string() before regexp_replace`, `90b2126 add quantile functions …`, `9d4878b Silence expected SAWarning …`.

On **pure master**, `test_sqlalchemy_functions.py` is **64/64 green**. After merging current master into this branch (conflict-free textually), it goes **red** — this PR's 3252-line parallel `starrocks_compat.py` registers `@compiles` handlers globally on import and overrides master's now-canonical StarRocks compilation, breaking `TestImportColStarrocks` (and, under `--doctest-modules`, cascading into other suites). rad-pat's approval here (2026-03-27) predates *all* of the above master work and no longer reflects reality.

Reconciling a 3252-line generated compat layer against master's now-authoritative implementation is not worth it when master already covers the need.

**Salvage probe (function-name surface):**
- **22 overlap / now conflicting** (master is authoritative — do not re-add): `to_char, to_date, to_number, to_timestamp, unix_to_timestamp, transaction_timestamp, strpos, string_to_array, ltrim, rtrim, zfill, slice_string, numericize, integerize_round, integerize_truncate, metric_multiply, safe_divide, normalize_whitespace, quantile_cont, quantile_disc, quantile_tdigest, quantile_tdigest_weighted`.
- **~189 names defined here but not in master's `sqlalchemy_functions.py`** — *potential* follow-up only. This is largely speculative breadth from generation; it should be filtered to functions **actually used in the codebase and currently unsupported on StarRocks** before any are re-implemented. Rough groups: `array_*` (~24), `json_*`/`json_path_*` (~20), `to_int*/to_uint*/to_float*` casts (~12), `to_start_of_*` time-bucketing (~14), date helpers (`add_months, months_between, next_day, to_year/to_month/...`), and misc (`iff, nvl, nvl2, decode, translate, split_part, regexp_*`, hashing `sha/xxhash*/siphash*`, etc.).

Recommendation: if any specific function above is genuinely needed on StarRocks and master lacks it, open a small targeted PR adding just that one to `sqlalchemy_functions.py` (master's pattern), rather than resurrecting this layer. Closed as superseded.

PR #293, opened 2026-05-15
Lazy load connection properties

rad-pat:
## Summary

- **Lazy-load** `variables` and `udf` on `Connection`, deferring RPC calls until the property is first read — reduces RPC call count for short-lived connections that never need them.
- **Revert** `ac62dad` (`fix(import_col): wrap col in to_string() before regexp_replace`). The `import_col` expression is only intended for the text-column import path. The original "bug" being fixed was that the code path was being incorrectly accessed with data from a parquet file, which should never have happened. That access has been corrected upstream, so the `to_string()` wrap is unnecessary and the original `import_col` behaviour should be restored.
- **Behavior fix in `query.py` `to_bool`**: strings in `falsey_strings` (`'f'`, `'F'`, `'no'`, `'false'`, `'FALSE'`) were converting to `True`. Now correctly convert to `False`. Callers that were relying on the previous (broken) semantics when reading Boolean columns from CSV will see a behavior change. Covered by `test_falsey_string_converts_to_false`.
- **`udf` property cleanup** (follow-up from review): `_load_udf_params()` now returns `UDFParams | None` instead of using a setter-as-loader pattern, removing an unreachable branch in the property.

## Test plan

- [x] Existing unit tests pass (`pytest plaidcloud/utilities/tests/test_query.py`).
- [ ] Verify in a downstream consumer that lazy-loading does not break callers that read `conn.variables` / `conn.udf`.
- [ ] Verify upstream parquet/bundle-import path no longer reaches `import_col` with non-text columns.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/pr_lint.yaml
+++ b/.github/workflows/pr_lint.yaml
@@ -14,22 +14,29 @@ jobs:
           python-version: '3.12'
       - name: Checkout code
         uses: actions/checkout@v6
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+        with:
+          # Install a specific version of uv.
+          version: "0.11.10"
       - name: Install dependencies
-        run: pip install -e .[test]; pip install -q pylint==${{ secrets.PYLINT_VERSION }};
+        run: |
+          uv pip install --system -e .[test]
+          uv pip install --system -q ruff
       - name: License Checker
         uses: andersy005/gh-action-py-liccheck@main
         with:
           strategy-ini-file: ./liccheck.ini
           level: cautious
           requirements-txt-file: ./requirements.txt
-          liccheck-version: ${{ secrets.LICCHECK_VERSION }}
+          liccheck-version: ${{ vars.LICCHECK_VERSION }}
       - name: Install ReviewDog
         uses: reviewdog/action-setup@v1
       - name: Perform Lint
         env:
           REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          pylint -s n -f text plaidcloud 2>&1 | reviewdog -efm="%f:%l:%c: %m" -name="PyLint" -reporter=github-check -level=warning
+          ruff check plaidcloud --output-format=concise --exit-zero 2>&1 | reviewdog -efm="%f:%l:%c: %m" -name="Ruff" -reporter=github-check -level=warning
       - name: Perform Tests
         run: |
           pytest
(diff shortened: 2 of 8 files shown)


rad-pat:
@claude review

rad-pat:
@inviscid, claude review failed here

simozzy:
1. The revert in eec62a5 is unexplained and undoes a documented production fix.
The reverted commit's message said regexp_replace(Decimal, ...) was being rejected by databend with no function matches signature, and that the to_string() wrap was added specifically for the parquet/bundle-import path where the source column keeps its native (non-text) dtype. The revert silently re-introduces that failure mode. Either:

Explain in the commit/PR description what made the wrap unnecessary (was the underlying problem fixed elsewhere? — e.g., bundle import now casts to text first?), or
Don't merge the revert.
This is the single most important thing to clear up before merging.

2. A real behavior fix is buried inside "Add tests and fix issues".
In [query.py:308](vscode-webview://1bbp613enmmacl1pifmlu5udpu60im50kugoqjigoapn69n4cfgp/plaidcloud/utilities/query.py#L308):


def to_bool(val):
    if val in falsey_strings:
-       return True
+       return False
    return bool(val)
Strings like 'f', 'no', 'false', 'FALSE' used to convert to True when reading Boolean columns from CSV. That's plainly a bug — but it's also a behavior change that any caller depending on the old (broken) semantics will silently flip on. It deserves its own commit and a callout in the PR body, not a stealth fix inside a test-coverage commit. The test at test_falsey_string_converts_to_false covers it, which is good.

3. The udf property has dead code and an awkward setter-as-loader pattern.


@property
def udf(self):
    if self._udf is self._NOT_LOADED:
        try:
            self._load_udf_params()
        except Exception:
            self._udf = None
        if self._udf is self._NOT_LOADED:   # ← dead: _load_udf_params always sets self.udf
            self._udf = None
    return self._udf
_load_udf_params() starts with self.udf = None ([query.py:735](vscode-webview://1bbp613enmmacl1pifmlu5udpu60im50kugoqjigoapn69n4cfgp/plaidcloud/utilities/query.py#L735)), so after the try succeeds _udf cannot be _NOT_LOADED. The fallback is unreachable. Cleaner pattern — return the value, set in one place:


def _load_udf_params(self) -> UDFParams | None:
    if not isinstance(self.rpc.step_id, str):
        return None
    config = self.rpc.analyze.step.step(...)
    ...
    return UDFParams(...)

@property
def udf(self):
    if self._udf is self._NOT_LOADED:
        try:
            self._udf = self._load_udf_params()
        except Exception:
            self._udf = None
    return self._udf

rad-pat:
Thanks for the review @simozzy — addressed all three points:

**1. `import_col` revert (eec62a5).** The `import_col` expression was only ever built to deal with text columns as part of an import file scenario. The previous `to_string()` fix was masking a different bug: the code path was being incorrectly accessed with data from a parquet file, which should never have happened. That upstream access has been corrected, so the `to_string()` wrap is unnecessary and the original `import_col` behaviour is the correct one. Updated the PR description to call this out.

**2. `to_bool` behavior fix.** Agreed it deserved a callout — I've added it to the PR description as a separate, prominent bullet rather than leaving it implicit in the test-coverage commit. Any caller depending on the old (broken) semantics will see strings like `'f'`, `'no'`, `'false'` flip from `True` to `False`.

**3. `udf` property dead code.** Refactored in f9d0d76 — `_load_udf_params()` now returns `UDFParams | None` and the property assigns `self._udf` in one place. Removed the unreachable `if self._udf is self._NOT_LOADED` branch. Existing tests still pass.

simozzy (approved):

PR #296, opened 2026-05-19
fix get_data type annotations

chucklay:


rad-pat (approved):

plaidcloud/utilities/query.py
```diff
@@ -203,7 +203,7 @@ def _csv_stream(self, file_name, columns, preserve_nulls):
 
                 yield row
 
-    def get_data(self, data_source: "Table|str", return_type='df', encoding='utf-8', clean=True):
+    def get_data(self, data_source: "Table|str|sqlalchemy.sql.Select", return_type='df', encoding='utf-8', clean=True) -> "str|pd.DataFrame":
```
inviscid, line 206: Accurate widening — this matches the `isinstance(..., sqlalchemy.sql.Select)` branch just below, and the return union is right (DataFrame for `df`, file-path `str` for `csv`).

One caveat the annotation can't express: a `Select` is only valid when `return_type='df'`. In the `csv` branch (`return_type='csv'`) a `Select` isn't handled and falls through to `raise Exception('Unknown type for Data Source ...')`. Not blocking — the union is still strictly more accurate than `"Table|str"` — just flagging that `get_data(some_select, return_type='csv')` will raise rather than work.

inviscid (approved):
Correct, low-risk annotation fix. The widened parameter matches the existing `isinstance(data_source, sqlalchemy.sql.Select)` handling, the new `-> "str|pd.DataFrame"` return matches both paths (DataFrame in `df` mode, file-path `str` in `csv` mode), and `pd`/`sqlalchemy` are both imported so the string annotations resolve for type checkers. One non-blocking note inline. LGTM.

_(Manual review via Claude Code, not the automated reviewer.)_

PR #305, opened 2026-06-06
Decode Alteryx SpatialObj fields to WKT in yxdb_to_csv

inviscid:
## What (Issue G)
A converted Alteryx workflow that imports a `.yxdb` with a spatial column (e.g. `05 Advanced/02 Map a trade area around a store`) failed downstream in spatial SQL: `Invalid WKT ... st_geometryfromwkt('b\x05\x00\x00\x00...')`. `yxdb_to_csv` read the `SpatialObj` field back as a raw binary blob (Alteryx's shapefile-shape layout) and wrote it into the CSV verbatim, so the imported column wasn't geometry the rest of the pipeline could parse.

## Fix
The `yxdb` library already decodes these blobs to GeoJSON (`yxdb.spatial.to_geojson`). Convert that GeoJSON to **WKT** so `.yxdb` geometry reads the same way as `.TAB`/executor geometry — which the spatial converter consumes via `st_geometryfromwkt` (CreatePoints/TradeArea/SpatialMatch/Distance). A small `_geojson_to_wkt` helper (Point/MultiPoint/Line/MultiLine/Polygon/MultiPolygon) avoids adding a shapely dependency.

Decode is **gated on the field's raw Alteryx type being `"SpatialObj"`** (read from the reader's metainfo, matching the `yxdb` lib's own `Blob`/`SpatialObj` distinction) — not on "the value is `bytes`". A generic `Blob` column is therefore never handed to the shape parser: arbitrary bytes whose leading bytes look like a shape type can make `to_geojson` raise (`struct.error`) or spin on a bogus point count. A malformed `SpatialObj` value also degrades to its raw value rather than aborting the whole import. Non-spatial files are byte-for-byte unchanged.

## Tests (pass locally on py3.11 with the real `yxdb` lib)
- `_geojson_to_wkt` for Point and Polygon
- a SpatialObj blob → WKT roundtrip via the real `yxdb.spatial.to_geojson`

## Release / rollout
plaid-utilities ships via a tagged GitHub Release → PyPI; after merge it needs a version bump + release, then a pin bump in plaid (`requirements.txt`) and workflow-runner (`requirements-base.txt`), then the bf2 image redeploy. Then re-run `02 Map a trade area` to confirm `.yxdb` geometry flows through the spatial chain.

### DIFF
--- a/plaidcloud/utilities/frame_manager.py
+++ b/plaidcloud/utilities/frame_manager.py
@@ -2464,6 +2464,37 @@ def parquet_to_csv(parquet_file_name: str, csv_file_name: str, start_row: int = 
 
 
 
+def _geojson_to_wkt(geojson: dict) -> str:
+    """Minimal GeoJSON -> WKT for Alteryx SpatialObj geometry.
+
+    Avoids a shapely dependency. Handles the geometry types the ``yxdb`` library's
+    ``spatial.to_geojson`` emits (Point, MultiPoint, LineString, MultiLineString,
+    Polygon, MultiPolygon).
+    """
+    geom_type = geojson.get('type')
+    coords = geojson.get('coordinates')
+
+    def ring(points):
+        return '(' + ', '.join(f'{x} {y}' for x, y in points) + ')'
+
+    if geom_type == 'Point':
+        return f'POINT ({coords[0]} {coords[1]})'
+    if geom_type == 'MultiPoint':
+        # OGC/ISO form parenthesizes each point -- strict WKT parsers require it.
+        return 'MULTIPOINT (' + ', '.join(f'({x} {y})' for x, y in coords) + ')'
+    if geom_type == 'LineString':
+        return 'LINESTRING ' + ring(coords)
+    if geom_type == 'MultiLineString':
+        return 'MULTILINESTRING (' + ', '.join(ring(line) for line in coords) + ')'
+    if geom_type == 'Polygon':
+        return 'POLYGON (' + ', '.join(ring(r) for r in coords) + ')'
+    if geom_type == 'MultiPolygon':
+        return 'MULTIPOLYGON (' + ', '.join(
+            '(' + ', '.join(ring(r) for r in poly) + ')' for poly in coords
+        ) + ')'
+    raise ValueError(f'Unsupported GeoJSON geometry type: {geom_type!r}')
+
+
 def yxdb_to_csv(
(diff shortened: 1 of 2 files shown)


inviscid:
## Adversarial review trail (self-run sonnet passes)

This PR went through three independent adversarial sonnet reviews before/while it was ready. Summary for the record:

**Pass 1 — caught a real bug (fixed):** the first cut only caught `TypeError`, but `yxdb.spatial.to_geojson` on a *non-spatial* `Blob` whose leading bytes resemble a shape type can raise `struct.error` (crash) or spin on a bogus point count (hang). Also a dead `if not geojson` guard.
→ **Fix:** decode is now gated on the field's raw Alteryx type being `"SpatialObj"` (read from the reader's metainfo, matching the yxdb lib's own `Blob`/`SpatialObj` check at `_yxdb_record.py:93`) rather than "value is bytes", so a generic `Blob` is never handed to the shape parser at all; plus a broad `except` so a malformed spatial value degrades to its raw value. Dead guard removed.

**Pass 2 — clean** on the gated approach: verified `reader._fields` is populated before read, `"SpatialObj"` is the exact constant, non-spatial files are byte-for-byte unchanged, and the hang/OOM path is eliminated for generic blobs.

**Pass 3 — caught a correctness nit (fixed):** `MULTIPOINT` was emitting the non-OGC form `MULTIPOINT (x y, x y)`; strict WKT parsers (PostGIS strict, some databend builds) require `MULTIPOINT ((x y), (x y))`.
→ **Fix:** parenthesized per OGC/ISO. Also strengthened the SpatialObj→WKT roundtrip test from `startswith` to full equality.

**Pass 4 — clean, merge-ready.** No correctness issues or regressions. Re-verified WKT validity for all six geometry types (incl. MultiPolygon ring nesting), the type gate + safe degradation, `orjson.loads(str)` + float fidelity, and the 3 unit tests (pass on py3.11 with the real `yxdb` lib).

**Known non-blockers:** no full `.yxdb` *fixture* with a spatial column exists, so the unit tests cover `_geojson_to_wkt` + the `to_geojson`→WKT roundtrip rather than the whole `yxdb_to_csv` pipeline (no real fixture available). End-to-end validation is the post-merge rollout: release → pin bump in plaid/wfr → bf2 redeploy → re-run `02 Map a trade area` to confirm `.yxdb` SpatialObj flows as WKT through the spatial SQL chain.

PR #310, opened 2026-06-17
fix(security): block eval() sandbox escape in expression engine [sc-22664]

simozzy:
## Severity: CRITICAL — remote code execution

`eval_expression` / `eval_rule` compile and `eval()` user-authored expressions in a namespace with **no `__builtins__` key**. CPython then auto-injects the full real builtins, and there was no AST allowlist on this path — so a crafted expression runs arbitrary Python/shell in the app and workflow-runner processes.

### Confirmed (benign localhost PoC through the product validator)
Fed `__import__('os').popen('id').read()` as a transform-step column expression to the exact save-time validator (`get_simple_expression_sql` → `get_select_query` → `eval_expression`). `id` executed **and the validator returned success** — the malicious expression validates clean while running code. Runtime user is in the `docker` group → host-root reachable.

### Reachability (authenticated write)
Transform-step save validation (`transform.py:278`), `model_handler.py:988`, Table Explorer, and the workflow runner — via any `target_columns[].expression`, `source_where`, or `having`. `step_upsert` (MCP) is not on the catastrophic denylist, so a prompt-injected agent with step-write can plant it; it fires at validate time, no run required.

### Context
`data-explorer-service` previously sandboxed this `compile()/eval()`; retiring it moved the eval in-process. Per-tenant namespaces bound cross-tenant blast radius but not in-tenant escalation or prompt-injection.

## Fix
- `get_safe_dict` pins `__builtins__` to `{}`.
- New `assert_eval_safe` AST gate at **both** eval sites rejects dunder name/attribute access (blocks `__import__`, `__builtins__`, and `().__class__.__bases__[0].__subclasses__()`) plus a denylist of dangerous builtins (`eval/exec/open/getattr/globals/...`).
- The expression language is SQLAlchemy (`func.*`, `cast`, type constants), so removing Python builtins is non-breaking.

## Verification
- RCE payload + `getattr`/`eval`/`open` + subclass-walk → all raise `SQLExpressionError`.
- 5 legitimate expressions still compile (`table.amount * 2`, `func.coalesce`, `cast`, `get_column`, `and_`).
- New `TestEvalSandbox` (5 tests) passes.
- No regression: `test_sql_expression.py` is 87 failed / 46 passed both before and after (the 87 are pre-existing dialect skew in the local env, unrelated to this change).

## Follow-ups (separate tickets, lower severity — from the same audit)
MCP surface un-rate-limited; `query_run` unbounded `max_rows` (OOM); `connection_describe` leaks DB passwords to the model; chat `run_query` weaker SQL guard; no orchestrator recursion/cost budget; audit-log gaps on credential/ACL mutations.

### DIFF
--- a/plaidcloud/utilities/sql_expression.py
+++ b/plaidcloud/utilities/sql_expression.py
@@ -3,6 +3,7 @@
 
 """Utility library for sqlalchemy metaprogramming used in analyze transforms"""
 
+import ast
 import re
 import uuid
 from copy import deepcopy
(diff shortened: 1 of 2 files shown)


simozzy:
Closing in favour of **#328**, which supersedes this. Same vulnerability (sc-22664), same file, same eval RCE — but #328 takes the fix further where it matters.

**Why this approach wasn't enough.** This PR's `assert_eval_safe` is a **denylist**: it blocks dunder attribute/name access and a list of dangerous builtins, and pins `__builtins__={}`. An 8-lens adversarial review on #328 found **26 confirmed executing bypasses** that a denylist structurally cannot catch — the whole `sqlalchemy` module is a launchpad to `os`/`subprocess`/`sys.modules` through **non-dunder** attribute chains. Run this PR's guard verbatim and all of these pass it:

```
sqlalchemy.log.logging.os.system('id')      # → RCE
sqlalchemy.text('DROP TABLE t')             # → raw-SQL injection
'{0.__class__}'.format(())                  # → str.format getattr, dunder hidden in the literal
```

None contain a dunder attribute or a forbidden name, so the denylist waves them through.

**What #328 does instead.** A strict **allowlist**: only curated AST node types are permitted, and any attribute chain rooted at `sqlalchemy` is confined to type/expression helpers — which closes the module-re-export traversal *and* `sqlalchemy.text` in one rule. It also keeps the `__builtins__` pin (curated safe set rather than empty, so `abs`/`str`/etc. still work) and blocks `str.format`. Verified against the full attack corpus + a no-regression gate (existing `test_sql_expression.py` matches the master baseline exactly). The adversarial review also surfaced one *separate* finding this PR didn't reach — pre-guard code exec via `apply_variables` on malicious variable *values* — tracked as sc-23058.

Nothing here is lost: the test cases are a subset of #328's corpus, and `assert_eval_safe` isn't imported anywhere (neither PR merged).

PR #311, opened 2026-06-17
fix(query): flatten compiled SQL with a space, not by deleting newlines

simozzy:
## What

`str(compiled_query).replace('\n', '')` flattened compiled SQL to one line by **deleting** newlines. When a statement carries a newline with no trailing space — as raw multi-line SQL embedded via `sqlalchemy.text()` does (an AI- or hand-authored extract) — that welds tokens across clause boundaries:

```
"driver_name"FROM ...      rJOIN ...      "year"AND ...
```

…which the warehouse rejects with a syntax error. Normal sqlalchemy-rendered SQL happens to emit a trailing space before each newline, so it dodged the bug; embedded raw SQL does not.

Replacing the newline with a **space** keeps the single-line form valid in both cases.

## Where

- `Connection._compiled` (`query.py`) — the live RPC query path
- module-level `compiled()` (`analyze_table.py`) — same one-liner

## Test

Added `TestCompiled.test_compile_flattens_embedded_newlines_with_space`: compiles a `text()` whose newline has no trailing space and asserts the flattened output keeps `"driver_name" FROM` (not welded). Fails on the old `replace('\n', '')`.

## Context

Surfaced while building the AI "SQL Extract from prompt" feature in `plaid` (sc-21457), which generates pretty-printed multi-line SELECTs. The plaid side has a flatten-safe workaround (trailing space per line); this fixes the root cause so any hand-written multi-line extract SQL is safe too.

### DIFF
--- a/plaidcloud/utilities/analyze_table.py
+++ b/plaidcloud/utilities/analyze_table.py
@@ -143,7 +143,10 @@ def compiled(sa_query, dialect='greenplum'):
     dialect = dialect or 'greenplum'  # Just in case someone sends a blank string, or a None by mistake
     eng = sqlalchemy.create_engine(f'{dialect}://127.0.0.1/', paramstyle='pyformat')
     compiled_query = sa_query.compile(dialect=eng.dialect, compile_kwargs={"render_postcompile": True})
-    return str(compiled_query).replace('\n', ''), compiled_query.params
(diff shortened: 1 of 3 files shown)


rad-pat (approved):
Approve, but too many comments for a quick fix. Chill out Claude!

PR #316, opened 2026-07-06
perf(query): cache project schema + drop redundant Table.__new__ touch (sc-22839)

rad-pat:
## Summary

Part of **sc-22839** task 3 (reduce `allocation_dim` step run time). This is the low-risk, **platform-wide** piece: it trims the per-`Table` metadata-RPC tax that dominates fixed step overhead. `Table.__new__` runs ~10-15 times per `allocation_dim` step (and constantly everywhere else), and was paying two avoidable RPCs each time.

### 1. Cache the project schema on the `Connection`
`Table.__new__` called `analyze.project.get_project_schema` on every instantiation. The physical schema is stable for the life of a connection, so it's now resolved once via a lazily-cached `Connection.project_schema` property (same `_NOT_LOADED` pattern as the existing `variables`/`udf` caches). Non-`Connection` callers fall back to the original inline RPC.

### 2. Drop the redundant write-path `touch`
The two `table.touch` calls in `Table.__new__` were **not** a trivial duplicate — line 843 reassigns `columns` from the input meta to the reflected `table_meta` between them, and `get_table` (no input columns) only ever reached the second. Server-side, `touch(overwrite=True)` does a **full recreate + `update_shape`**, so on every `save_data`-style call the second touch recreated the just-created table again and rebuilt reporting views a second time.

It's now gated on `not had_input_columns`:
- **write path** (columns supplied) → one touch, no redundant recreate;
- **read path** (`get_table`) → the ensure-exists touch is preserved unchanged.

## Tests
Added 3 cases to `TestTable` pinning the new contracts (write path touches once, read path still ensures once, schema resolved once per connection). Full suite green: **468 passed, 4 pre-existing skips**.

## Rollout note
Shared-lib change — it only reaches running workflow steps once a new plaid-utilities version is published and the pins in `workflow-runner/requirements-base.txt` and plaid's `requirements.txt` are bumped.

## Not included
The larger `allocation_dim`-specific win (collapsing the allocable→paths→mapping recreate/read-back chain with an inline `path_id`) is a deliberately separate, riskier follow-up.

### DIFF
--- a/plaidcloud/utilities/query.py
+++ b/plaidcloud/utilities/query.py
@@ -96,6 +96,7 @@ class Connection:
         self.dialect = dialect_cls(paramstyle='pyformat')
         self._variables = self._NOT_LOADED
         self._udf = self._NOT_LOADED
+        self._project_schema = self._NOT_LOADED
 
     @property
     def variables(self):
@@ -730,6 +731,18 @@ class Connection:
             raise Exception('Project Id has not been set')
         return self._project_name
 
+    @property
+    def project_schema(self):
+        # The physical schema is stable for the life of the connection, so cache
+        # it rather than paying an RPC on every Table instantiation (there are
+        # ~10-15 per allocation step).
+        if self._project_schema is self._NOT_LOADED:
+            if self._project_id.startswith(SCHEMA_PREFIX):
+                self._project_schema = self._project_id
+            else:
+                self._project_schema = self.rpc.analyze.project.get_project_schema(project_id=self._project_id)
+        return self._project_schema
+
     def get_dimension(self, dimension_name):
         return self.dims.get_dimension(name=dimension_name, replace=False).hierarchy_table()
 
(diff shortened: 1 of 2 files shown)


simozzy (commented):
**Review — sc-22839: cache project schema + drop redundant `Table.__new__` touch**

Traced both changes against the installed source. Correct, well-tested, low-risk. Approve.

---

**1. `Connection.project_schema` cache — ✅ safe**

- `_project_id` is assigned only in `Connection.__init__` — there's no setter and no `set_project`, so it's immutable for the connection's life. Caching the derived schema is therefore correct, and it mirrors the existing `variables`/`udf` `_NOT_LOADED` caches exactly.
- The property reproduces the inline logic faithfully (`SCHEMA_PREFIX` short-circuit → project_id, else the same `get_project_schema` RPC), so the resolved value is identical — just fetched once.
- The `isinstance(conn, Connection)` guard is right: the cached property lives on `Connection`, and non-`Connection` callers keep the original inline branch. No behaviour change for them.

**2. Dropping the second `touch` — ✅ correct, and lower-risk than it first looks**

The reassurance that made me comfortable: **the returned SQLAlchemy `Table` object is unaffected by this change.** `super().__new__` is built from the *reflected* `columns` (post-`table_meta`) and `_schema` regardless of the touch gating, and those lines are untouched. So the only thing elided is the redundant server-side second `touch` (full recreate + `update_shape`) — the Python-side object callers receive is byte-for-byte the same.

- `had_input_columns = bool(columns)` is captured after touch1 but before `table_meta` reassigns `columns` — correct placement, and its truthiness tracks touch1's own `if columns:` guard, so "input columns supplied ⇒ touch1 definitely ran" holds.
- Read path (`get_table`, no input columns) → `had_input_columns` False → the ensure-exists touch still runs. Preserved.
- Empty-list edge (`columns = []` on the new-table branch) is falsy, so both the write-path skip and the no-op cases behave as before.

**One assumption worth stating** (not a blocker — you've verified it server-side): on the write path the physical table is now left in the shape of the **caller's input columns** (touch1) rather than the **reflected `table_meta`** (the old touch2). These are the same table's metadata reflected back, so they match unless the server canonicalises columns in `table_meta` differently from what `touch` writes. If that ever diverges, the physical result would differ — but the in-memory `Table` wouldn't, and your test pins the surviving touch's meta explicitly, which is the right guard.

**Tests — good.** Three cases pin exactly the three contracts that could regress (schema fetched once/connection, write-path single touch carrying input meta, read-path ensure preserved with reflected meta). That's the contract, not the implementation.

Rollout note is correct — inert until the pin is bumped in `workflow-runner/requirements-base.txt` and plaid's `requirements.txt`.

Verdict: **Approve.**

rad-pat:
@simozzy possible to approve so I can merge?

simozzy (approved):

PR #318, opened 2026-07-06
fix(frame_manager): .yxdb import data fidelity (CP1252, exact FixedDecimal, no bytes repr)

inviscid:
## What

Three data-corruption defects in the `.yxdb` → staged-CSV conversion that made Alteryx `.yxdb` imports silently lose or mangle data:

- **String/V_String → CP1252.** The `yxdb` library decodes narrow strings as latin-1, but Alteryx narrow strings are CP1252 in practice, so `0x80`–`0x9F` bytes (euro, curly quotes, en/em dashes) imported as C1 control characters. Now transcoded latin-1 → CP1252 post-decode (lossless byte round-trip; `errors='replace'` for the five undefined CP1252 bytes `0x81/0x8D/0x8F/0x90/0x9D`).
- **FixedDecimal → exact string.** The library floated the on-disk ASCII decimal string, destroying digits past ~15–16 significant figures before the Decimal-typed column ever saw the value. The library's extractor factory is now overridden (scoped to reader construction) so the exact decimal string flows through the CSV to the typed cast.
- **Blob / unparseable SpatialObj → hex.** Bytes values reached `DataFrame.to_csv` and were written as the Python `b'...'` repr. They now land as hex; parseable SpatialObj blobs keep decoding to WKT as before.

Conversion is factored into `_convert_yxdb_value` so the fidelity rules are unit-testable without a real `.yxdb` binary; an end-to-end test runs a synthetic `.yxdb` (FixedDecimal + V_String + Blob) through `yxdb_to_csv`.

## Testing

`plaidcloud/utilities/tests/test_yxdb_to_csv.py` — 13 passed (unit + synthetic end-to-end). Full suite green.

## Deploy note

⚠️ `plaid-utilities` is an installed pip dependency, not a live service. Merging this to `master` does **not** deploy it — it requires a subsequent release tag + a version-pin bump in the consuming repos (plaid / workflow-runner). **Do not auto-cut the release.** This PR lands the fix; the release is a separate, deliberate step.


inviscid (commented):

Independent Sonnet review verified all three .yxdb fidelity fixes against the actual `yxdb` PyPI package source (1.1.1): the latin-1→CP1252 re-transcode is a lossless round-trip and the 5 undefined bytes (0x81/0x8D/0x8F/0x90/0x9D) are exactly those Python's cp1252 codec rejects; the FixedDecimal extractor override correctly mirrors the library's null-flag/termination logic and is looked up via a module-attribute reference so the scoped monkeypatch genuinely takes effect; the only production caller uses ProcessPoolExecutor (separate processes), so the documented thread-safety caveat is real-but-latent, not an active bug. Blob/SpatialObj hex/WKT handling and the synthetic end-to-end test exercise corrected values, not mocked assertions. Non-blocking notes (untyped-fallback branch lacks a dedicated test; reverts to already-shipped behavior) documented; accidentally-committed coverage artifacts have been stripped from the branch.

PR #319, opened 2026-07-07
Detect AMP-engine (.yxdb e2) files and fail with an actionable message

inviscid:
## Summary
- `yxdb_to_csv` now checks the file header before constructing `YxdbReader`: files written by Alteryx's AMP engine (header `Alteryx e2 Database file`, unpublished spec that no library can read) raise a clear `ValueError` instead of the yxdb library's cryptic `IOError`.
- The error tells the user exactly how to recover: re-save in Alteryx via the Output Data tool's "compatible with Designer 18.1 and older" option (AMP reads and writes both formats), or export to CSV and re-import.

## Tests
- New test: a synthetic file with the e2 header raises `ValueError` mentioning AMP.
- Full `test_yxdb_to_csv.py` suite: 15 passed (all existing classic-format tests unaffected).

### DIFF
--- a/plaidcloud/utilities/frame_manager.py
+++ b/plaidcloud/utilities/frame_manager.py
@@ -2571,6 +2571,8 @@ def yxdb_to_csv(
 
     Raises:
         :class:`ImportError`: if the ``yxdb`` package is not installed.
+        :class:`ValueError`: if the file is in the unsupported AMP-engine
+            ("Alteryx e2 Database file") format.
     """
 
     try:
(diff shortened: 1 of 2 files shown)


inviscid (commented):

Verified against the yxdb 1.1.1 wheel source: the sniff uses its own handle (YxdbReader reopens by path), the classic 21-byte magic cannot collide with the e2 prefix (no false positives), the message is accurate and actionable, and the test triggers the guard correctly. No strays.

PR #323, opened 2026-07-08
Make generated SQLAlchemy functions pickleable

inviscid:
## Summary
- bind dynamically generated StarRocks SQLAlchemy function classes into the module namespace
- add regression coverage that every generated function class can round-trip through pickle

## Why
Workflow query serialization pickles SQLAlchemy expressions. The generated functions such as `to_year` and `to_month` registered with SQLAlchemy but were not available as `plaidcloud.utilities.sqlalchemy_functions.<name>`, so pickle could not resolve them after the Unio upgrade.

## Validation
- `/Users/inviscid/Projects/plaid-utilities/.venv/bin/pytest /private/tmp/plaid-utils-pickle-fix/plaidcloud/utilities/tests/test_sqlalchemy_functions.py -q` (99 passed)


inviscid:
Self-review: approved. I reviewed the focused two-file diff and only found polish-level concerns. The generated SQLAlchemy classes are now module-resolvable for pickle, and the regression covers every generated rename.

PR #324, opened 2026-07-09
Compile keyword date_add for StarRocks

inviscid:
## Summary
- add a StarRocks compiler for keyword-based `date_add` expressions
- render keyword units as nested StarRocks `*_add` calls
- add focused coverage for weeks-plus-days expressions and no-op date_add

## Why
The existing generic `date_add(..., weeks=..., days=...)` compiler emits `make_interval`, which is not StarRocks SQL. This keeps the existing Databend/default rendering unchanged and adds only the StarRocks specialization.

This is not the Unio Databend dashboard fix; Unio uses Databend and needs its saved dashboard/filter SQL corrected separately.

## Validation
- `/Users/inviscid/Projects/plaid-utilities/.venv/bin/python -m pytest plaidcloud/utilities/tests/test_sqlalchemy_functions.py -q` (102 passed)
- `git diff --check`


inviscid:
Self-review approved. Scope is limited to StarRocks keyword date_add compilation; Databend/default rendering is unchanged. Checks passed.

PR #328, opened 2026-07-15
fix(security): sandbox the expression engine against eval RCE (sc-22664)

simozzy:
## Problem (sc-22664)

`eval_expression` / `eval_rule` compile and `eval()` user-authored strings to build SQLAlchemy expression trees. The eval globals (`get_safe_dict`) carried **no `__builtins__` key**, so CPython injected the *real* builtins module. Anyone who can author a transform-step expression — calc field, filter/`where`, rule-based tagging, allocation — could therefore run arbitrary code:

```python
eval_expression("__import__('subprocess').getoutput('id')", None, [])
# → 'uid=1000(plaiddev) gid=1000(plaiddev) ...'   (confirmed live)
```

This is a **live, authenticated RCE**. Expressions execute in workflow-runner (which uses this module directly) and save-time validation triggers the same path, so per-customer namespace isolation only contains blast radius — it does not close the hole.

## The fix

Two independent layers, both wired through the single `get_safe_dict` chokepoint and **both** eval sites (`eval_expression` and `eval_rule`):

1. **Pinned `__builtins__`** — `get_safe_dict` sets `__builtins__` (last, so nothing merged in can reopen it) to a curated pure-function set (`abs round min max len sum str int bool`). Removes `__import__`/`open`/`eval`/`exec` from the eval context.
2. **`_assert_safe_expression`** — a static AST guard on the post-substitution string (exactly what gets eval'd). Rejects:
   - **dunder / private attribute access** — `().__class__.__mro__[1].__subclasses__()` reaches `os` with *no* builtins at all; a single leading underscore is allowed **only** as a bare SQL-function name off `func` (e.g. Databend's `func._if`).
   - **`str.format` / `format_map`** — `"{0.__class__.__init__.__globals__}".format(())` hides dunders inside a string literal the AST can't see; `func.format` (a real SQL `FORMAT`) stays allowed.
   - **lambdas and comprehensions** — no legitimate use in a SQL expression, and they widen the escape surface.

A `SyntaxError` is left for `compile()` to report as before, so error UX for malformed expressions is unchanged.

## Why this doesn't break real expressions

Attribute **roots** are dynamic here (table aliases, `extra_keys`), so unlike an allowlist-of-roots approach this guard never restricts roots — it only blocks the escape primitives, which SQL expressions never need.

- New `test_expression_sandbox.py`: 7 tests — every attack class blocked (import, traversal, format leak, lambda, comprehension), builtins pinned, and legit expressions (`func._if`, `func.format`, `cast`, `case`, `in_`, arithmetic) still pass and still evaluate.
- Swept **every** expression string in the existing test suite through the guard → **no false positives**.
- Ran the existing `test_sql_expression.py` against this branch vs. an unmodified-master baseline under identical conditions → **identical result set** (the environment's pre-existing `PlaidCurrency` drift errors are the same on both). **Zero regressions.**

## Downstream (not in this PR)

- Consumers pick this up only when a new utilities version is cut and **plaid bumps its pin** (`plaidcloud-utilities`, currently `==1.10.0`) and **workflow-runner** does the same — the latter is where transform expressions actually execute.
- **Out of scope, follow-up:** the raw-SQL surface (`sqlalchemy.text(...)` and the fully-exposed `sqlalchemy` module) remains reachable — that's SQL injection, not RCE, and restricting it risks breaking expressions that use `sqlalchemy.*`, so it's deliberately left for a separate change.


simozzy:
## Adversarial review outcome (8 attack lenses, run live against the module)

The first-pass **denylist was defeated with 26 confirmed *executing* bypasses**, so the guard was rewritten as a **strict allowlist** (commit `7a98f76`).

**What broke it (classes A + D):** the fully-exposed `sqlalchemy` module was a launchpad to `os`/`subprocess`/`io`/`sys.modules` through **non-underscore** attribute chains the denylist never inspected —
`sqlalchemy.log.logging.os.system(...)`, `sqlalchemy.util.concurrency.asyncio.base_events.subprocess.call(...)`, `sqlalchemy.engine.base.sys.modules['os']...`, `...inspect.importlib.import_module('os')...` — plus the same call wrapped in otherwise-allowed nodes (ternary / chained-compare / list+subscript / starred / bytes-decode). My earlier "leave `sqlalchemy.text` out of scope" note was wrong: the whole module, not just `text`, was reachable.

**Fix:** node-type allowlist + confine any `sqlalchemy`-rooted attribute chain to curated type/expression helpers. Closes `sqlalchemy.text` raw-SQL and every module-re-export in one rule.

**Found nothing (good signal):** Unicode/NFKC identifier tricks (guard sees post-normalization names), `eval_rule` parity (structurally identical protection), and parser-mode divergence (`ast.parse`/`compile` share the CPython parser on 3.13).

**Verification:** all 26 payloads now blocked & non-executing; the existing `test_sql_expression.py` result set is **identical** to the `origin/master` baseline (zero regressions); the legit-expression corpus still passes; regression tests added.

**Behaviour change worth knowing:** an allowlist is stricter than before — f-strings, dict/set literals, slices, and ternaries in expressions are now rejected. The full existing test suite passing identically is evidence these aren't used in practice, but flagging it.

---

## ⚠️ Separate, still-open finding (class F) — not fixed here

A `variables` dict whose **values are live Python objects** with a malicious `__str__`/`__repr__`/`__format__`/`__getitem__` execute inside `apply_variables`' `message.format(**variables)` — **before** this guard runs (the guard only ever sees the post-substitution string). Confirmed reproducing.

**Reachability is low:** it needs a *caller* to pass a live malicious object as a variable value. It is **not** reachable by authoring an expression string — production variables are config scalars. It lives in `stringtransforms.apply_variables` (a shared utility), so a fix belongs at the variable-handling boundary, not in `_assert_safe_expression`. Tracked as [sc-23058](https://app.shortcut.com/plaidcloud/story/23058).

rad-pat (approved):

PR #329, opened 2026-07-19
fix(security): allowlist sqlalchemy.select, block its raw-SQL splice …

simozzy:
…methods (sc-23125)

The sc-22664 sandbox confined a sqlalchemy-rooted chain to _ALLOWED_SQLALCHEMY_ATTRS, which omitted `select`. The Table Explorer facet's distinct-count is a runtime-generated eval'd expression (`sqlalchemy.select(func.count(func.distinct(get_column(table, <col>)))).subquery()`), so it was rejected with "Access to attribute 'select' is not permitted", the facet queries never ran, and the details/facets view never loaded.

Allowlist `select`, but close the vector it would otherwise open: a Select's prefix_with/suffix_with/with_hint/with_statement_hint splice a raw string literally into the compiled SQL (the same primitive as text()), and they live on the call RESULT — root None — so the sqlalchemy-root allowlist can't see them, nor can it see the `.element` back-reference from a subquery. Block those four method names unconditionally (mirroring the existing str.format/format_map block). Every other string-accepting Select method already requires text() in SA 2.0, which stays blocked.

Regression tests cover the facet happy-path and all five injection shapes (direct methods + subquery .element back-ref).


rad-pat (approved):
## Overview

The sc-22664 sandbox confined `sqlalchemy`-rooted attribute chains to `_ALLOWED_SQLALCHEMY_ATTRS`, which omitted `select` — breaking the Table Explorer facet's runtime-generated distinct-count expression. This PR allowlists `select` and adds `_RAW_SQL_METHODS`, blocking `prefix_with`/`suffix_with`/`with_hint`/`with_statement_hint` by name unconditionally.

I verified the claims empirically against SQLAlchemy 2.0.48 rather than taking the comments at face value.

## What checks out

- **The fix works.** All 5 attack shapes blocked, facet expression allowed, all 13 tests pass at PR head (`170df32`).
- **`_RAW_SQL_METHODS` is exactly complete.** Enumerating the mixins gives `HasPrefixes → {prefix_with}`, `HasSuffixes → {suffix_with}`, `HasHints → {with_hint, with_statement_hint}`. Four for four, nothing missed.
- **Blocking by *name* rather than by path is the right call.** It makes back-reference enumeration irrelevant — `.element` is one path back to the Select, but since the block is on the method name, any other back-ref (`.selectable`, `.original`, …) is covered for free. And it can't be bypassed via `getattr`, which isn't in `_SAFE_BUILTINS`.
- **The "requires `text()` in SA 2.0" claim holds.** `where`/`having`/`add_columns`/`column`/`with_only_columns`/`filter` reject plain strings with `ArgumentError`; `order_by`/`group_by`/`distinct` reject with `CompileError`. Worth noting the latter three fail at *compile* time, not build time — still safe (no SQL is ever emitted), but "already require text()" glosses over the deferred failure.

## Main comment: the stated invariant is broader than what's true

The code comment asserts select "cannot re-introduce a text()-style vector," and the PR body says the change closes that vector. That's true of `select` specifically — but **`literal_column`, already in `_ALLOWED_SQLALCHEMY_ATTRS`, is the same primitive.** Confirmed end-to-end through the real `eval_expression` path at PR head:

```
expr:   sqlalchemy.literal_column('(SELECT password FROM users LIMIT 1)')
result: SELECT (SELECT password FROM users LIMIT 1)
```

`literal_column` renders its string verbatim into compiled SQL — the exact capability `text()` is blocked for. It also composes through the newly-allowed `select` (e.g. inside `.where(...)`).

**This is pre-existing** — it arrived with sc-22664, not with this PR — so I don't think it should block the merge, and this PR is a net security improvement regardless. But two asks:

1. **Soften the comments** on `select` / `_RAW_SQL_METHODS` so they claim what's actually true ("`select` adds no new raw-SQL splice beyond `_RAW_SQL_METHODS`") rather than a sandbox-wide no-raw-SQL invariant a future reader might rely on when adding the next allowlist entry.
2. **File a follow-up ticket** for `literal_column`. It needs a real decision — it may well be load-bearing for legitimate expressions, in which case the answer is argument validation rather than removal.

## Minor

- **Name-based blocking is global.** A user column literally named `prefix_with`/`suffix_with`/`with_hint`/`with_statement_hint` becomes unreachable via `table.<name>`. I confirmed the escape hatch works — `get_column(table, 'suffix_with')` is allowed — so impact is negligible. Worth one line in the comment pointing at `get_column`, since the error message ("Access to attribute … is not permitted") would otherwise be baffling for that user.
- **Three near-identical raise blocks** now sit in the `ast.Attribute` handler with the same message string. Collapsing to a single name-check would tighten it, but the current shape matches the surrounding style — take it or leave it.

## Test coverage

Good, and follows the file's existing conventions (module-level attack list + `subTest` loops). Covers guard-level and eval-path level for all five shapes, plus the happy path asserting a real `Subquery`. The `.element` back-ref case is a nice touch.

One suggestion: since the security property is "these names are blocked *regardless of root*," a test asserting `func.prefix_with(...)` and `table.suffix_with` are also blocked would pin the intended global scope — right now that behavior is incidental rather than asserted.

---

*Note: the local `plaidcloud-rpc` in my environment predates `PlaidCurrency` (needs ≥ 1.9.0), so I shimmed that import to run the suite. Unrelated to this PR, but you'll hit it on a stale env.*

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

PR #332, opened 2026-07-21
sc-23158 WS-A2/A5: canonical GUIDHyphens uuid cast + databricks test deps

inviscid:
Part of [sc-23158](https://app.shortcut.com/plaidcloud/story/23158), plan items **WS-A2/A5**. Plan: [Shortcut Doc](https://app.shortcut.com/plaidcloud/write/IkRvYyI6I3V1aWQgIjZhNWU5NWRiLWE4Y2YtNGZiOC1hZTYxLTM5N2U2NTAxZmVkMCI=).

## What
- `uuid` dtype map: `postgresql.UUID` → dialect-neutral `sqlalchemy.Uuid` (verified zero behavior change on postgres; Databend goes from an invalid `CAST(x AS UUID)` to a working `CHAR(32)`).
- `SQLAlchemy>=2.0` floor declared (`sqlalchemy.Uuid` is 2.0-only; package ships to PlaidXL/custom-transform envs).
- databricks dialect deps added to the `[test]` extra.

## Validation
573 passed / 185 subtests; ruff clean on touched lines.

## Review
Adversarial review + fix cycle — APPROVED.


inviscid (commented):

Fixes a real data-integrity bug: the expression-language uuid/Uuid/UUID cast keys previously mapped to postgresql.UUID, producing invalid SQL on Databend and a silently-dropped CAST on StarRocks, and never matched the 36-char hyphenated CHAR(36) storage format that platform uuid-typed columns use via sqlalchemy_from_dtype('uuid') → GUIDHyphens. Verified the fix aligns the inline-cast path with the canonical column-dtype path; postgres behavior is byte-identical (GUID special-cases dialect.name == "postgresql" to native UUID). New multi-dialect tests assert real compiled SQL. Non-blocking note: the databricks test-extra additions are currently inert (no consuming test yet — arrives with later WS-B work).

PR #334, opened 2026-07-21
feat(connect): forward connection configuration to Connect (sc-21805)

inviscid:
Ref: [sc-21805](https://app.shortcut.com/plaidcloud/story/21805)

## Problem

`PlaidConnection.__init__` accepted `*args`/`**kwargs` and then called `Connect.__init__(self)` with **none of them**:

```python
def __init__(self, *args, **kwargs):
    ...
    Connect.__init__(self)      # nothing forwarded
```

Every argument was silently discarded. That made plaidcloud-rpc's direct-argument configuration — added in [plaid-rpc#158](https://github.com/PlaidCloud/plaid-rpc/pull/158) and released in [1.11.0](https://github.com/PlaidCloud/plaid-rpc/releases/tag/v1.11.0) — unreachable from the class that the generated UDF boilerplate actually constructs (`conn = PlaidConnection()`, `plaid/core/api_utilities/analyze/udf.py`). It is also why `platform-panel-base`'s `sitecustomize.py` monkeypatches this very constructor just to get a project id through.

## Change

Forward `rpc_uri`, `auth_token`, `token_provider`, `workspace_uuid`, plus `project_id` when `rpc_uri` is present.

Two deliberate details:

- **By explicit name, not `**kwargs`.** `xl_path` is consumed by `PlaidConnection` itself further down (`kwargs.get('xl_path')`) and `Connect` would raise `TypeError: unexpected keyword argument`. Blind forwarding breaks the xlwings path.
- **`project_id` only alongside `rpc_uri`.** Without `rpc_uri` the environment or `plaid.conf` supplies the project, and plaid-rpc 1.11.0 raises on a bare `project_id` rather than silently ignoring it.

Floor bumped to `plaidcloud-rpc>=1.11.0` in both `pyproject.toml` and `requirements.txt`.

## Backward compatibility

A bare `PlaidConnection()` forwards nothing and takes the same environment/`plaid.conf` path as before. The Jupyter branch still seeds its env vars unchanged.

## Tests

New `plaidcloud/utilities/tests/test_connect.py` — there was no test file for this module.

Verified the tests actually capture the fix rather than just passing: run against `origin/master` (pre-fix), the two forwarding tests **fail** and the five regression guards **pass**.

```
FAILED test_direct_config_reaches_connect
FAILED test_token_provider_reaches_connect
2 failed, 5 passed
```

Against this branch: 7 passed. No new ruff findings (the two in `connect.py` are pre-existing — the import block and the xlwings bare-except).

## Follow-up, not in this PR

With this landed, the workarounds this gap forced can be removed: Jupyter's `workflow_id_not_set` / `step_id_not_set` sentinels in this file, and `sitecustomize._install_panel_auth`'s placeholder env vars and two monkeypatches. Those need a `platform-panel-base` rebuild and are tracked on sc-21805.


inviscid:
Clean-room review came back **REQUEST_CHANGES** and found a real regression I introduced. Fixed in a997a37.

**The bug:** plaidcloud-rpc raises `ValueError('workspace_uuid and project_id require rpc_uri.')` when *either* is truthy without `rpc_uri` — but I gated only `project_id`, forwarding `workspace_uuid` whenever present. Callers have long passed whole plaid.conf-shaped blobs as kwargs and had the extras silently ignored:

```python
PlaidConnection(**yaml.safe_load(open('.plaid/plaid.conf')))   # carries workspace_uuid, no rpc_uri
```

That turned a working call into a hard `ValueError` — the opposite of this PR's backward-compatibility goal, and the exact failure the `project_id` gate already existed to prevent. Reproduced before fixing:

```
RAISES ValueError: workspace_uuid and project_id require rpc_uri.
```

**Fixes:**
- Gate `workspace_uuid` on `rpc_uri` too. Both now forward only on the direct path.
- Short-circuit the Jupyter branch when `rpc_uri` is given — it was seeding `__PLAID_PROJECT_ID__` and friends into `os.environ` even though the direct path never reads them back, leaking process-wide into any later bare `PlaidConnection()`.
- Two new tests for the regression (`workspace_uuid` alone; a plaid.conf-shaped blob) plus one for the Jupyter short-circuit. 10 tests, all passing; 3 of them fail against pre-fix `origin/master`.

Note on the review's evidence: it cited `udfs/demo/pwc/pwc_demo_01.py` and `udfs/saputo/.../rca_v32.py` as affected call sites. Those paths do not exist in the `udfs` checkout I have, so I could not confirm them — but the defect itself is real and reproducible independently of those files, which is what matters.

inviscid (commented):
✅ Clean Sonnet re-review of the corrected head (`a997a37`) — **APPROVE**

The first review returned REQUEST_CHANGES against `3065b97`; this is an independent re-review of the fix. It verified empirically rather than by reading: cloned the repo at `a997a37`, `pip install`ed real `plaidcloud-rpc==1.11.0`, ran the new tests against the actual package (10/10) rather than only the mocked fixture, and drove `PlaidConnection`/`PlaidConfig` directly through the flagged edge cases.

1. **Original bug fixed, no remaining leak.** `workspace_uuid`/`project_id` are written into `rpc_config` only inside `if rpc_config.get('rpc_uri'):`. Confirmed live against real 1.11.0 that `PlaidConfig(workspace_uuid=...)`, `PlaidConfig(rpc_uri='', workspace_uuid=...)` and `PlaidConfig(rpc_uri=None, workspace_uuid=...)` all raise, then confirmed those kwargs never reach it through `PlaidConnection` — the ValueError is unreachable.
2. **`elif is_jupyter` is correct and safe.** It only changes behaviour when `rpc_uri` is truthy — a path that could not be exercised before this PR, since nothing was forwarded at all. Matches plaid-rpc's documented precedence ("Passing rpc_uri selects the arguments and skips the other two entirely").
3. **Unconditional pop is safe** — `workspace_uuid` is not read anywhere else in `__init__`; only `xl_path` is consulted from the remaining kwargs.
4. **Empty-string forwarding is a no-op** — `PlaidConfig`'s own defaults for both are `''`, verified against source and by live call.
5. **Tests are contract-level and pass against the real package**, covering the regression (`workspace_uuid` alone, plaid.conf-shaped blob) and the Jupyter short-circuit.

A direct-config call with `rpc_uri`+`auth_token`+`workspace_uuid`+`project_id` clears all `PlaidConfig` validation and reaches an actual (expected) DNS failure — the fix is real, not test-shaped.

Two non-blocking nits, recorded rather than fixed since only nits remain:
- No test exercises `rpc_uri=''`/`rpc_uri=None` explicitly passed alongside `workspace_uuid`/`project_id`. Verified manually to behave correctly; coverage gap only.
- Under `__PLAID_JUPYTER__=True` with an explicit truthy `rpc_uri` and no `project_id`, the eager "Set the Project ID..." guard no longer fires; the failure defers to `PlaidConfig.project_id`'s own exception on first access. Intentional (avoids the `os.environ` leak) and covered by a test.

PR #335, opened 2026-07-21
fix(expressions): stop the sandbox rejecting `is`/`not` and underscore columns (sc-23186)

simozzy:
Fixes [sc-23186](https://app.shortcut.com/plaidcloud/story/23186).

## The regression

A customer on last night's release hit two failures on steps that had run for years. Both are false positives from the sc-22664 expression-sandbox hardening, and neither is what the error message suggests.

**1. `is` / `is not` / `not`**

```
ERROR   Error in expression:
    case(
        (get_column(table1, 'BillTo_Country') is None, false),
        ...
Is is not permitted in an expression.
```

`ast.Is`, `ast.IsNot` and `ast.Not` were simply omitted from `_ALLOWED_NODES`. They are pure operators — no attribute access, no call, no name resolution — so they add no escape surface. (`ast.IfExp` *is* a deliberate exclusion and stays excluded.)

**2. Underscore-led column names**

```
ERROR   Error in expression:
    func.rtrim(
    table._MajorAccountFlag
)
Access to attribute '_MajorAccountFlag' is not permitted.
```

Reported as "the system no longer likes `func.rtrim`" — it isn't `rtrim`, it's the column name. The guard blocked every underscore-led attribute as an internals-escape vector, but customer columns are legitimately named `_MajorAccountFlag`, and `table._MajorAccountFlag` is how the UI writes it.

## The fix, and the trap inside it

The three operators go back on the allowlist. The underscore-column exemption is the part worth reviewing carefully.

The naive version — allow the attribute if the name is in the table's column list — **is exploitable**, and I only found that by attacking my own first cut:

```
table._index                           PASS  -> builtins.dict          # ColumnCollection internal!
table._all_columns                     PASS  -> builtins.list
table._all_columns[0].table.metadata   PASS                            # reaches the schema graph
```

Naming a column `_index` does not *shadow* `ColumnCollection._index`; attribute access still returns the internal. So the exemption is keyed on **resolved identity**, not on the name:

```python
if getattr(collection, name, None) is collection[name]:
```

A column named after an internal is therefore left blocked and stays reachable only through `get_column(table, '_index')`, which cannot return internals. `__dunder__` names are never exempt.

Telling column from internal needs the eval namespace, so `_assert_safe_expression` now takes the `safe_dict` it is guarding; both call sites (`eval_expression`, `eval_rule`) already had it in scope. A caller that passes none keeps the old blanket block — closed by default.

## Restoring `not` reopens one silent-failure shape — which is rejected

Putting `ast.Not` back on the allowlist is not free: `not (<comparison>)` over a column does **not** build a SQL NOT. `==`/`!=` go through `BinaryExpression.__bool__` (an `id()`-based special case), `is`/`is not` are Python identity, `in`/`not in` bool() the per-element `==` — so the comparison collapses to a plain Python bool, the column drops out of the query, and the branch becomes a constant (`CASE WHEN true THEN ...`). That is *worse* than the loud failure every other `not` shape gives, and on the shipped release it could not happen at all because `ast.Not` was rejected outright. So the guard rejects the collapsing shapes and steers to what actually builds SQL:

- A single comparison → `!=`, `a.is_(None)`, `a.notin_([...])`, and so on.
- Comparisons joined by `and`/`or` (`not (a == 1 or b == 2)`) → `not_(and_(...))` / `not_(or_(...))`. The naive `~(a == 1 or b == 2)` does not work — the inner Python `or` collapses before `~` runs.

A `not (...)` wrapping a comparison of **pure constants** (no column — e.g. `not ('ABC' == 'CORP')`, or `not (len('abc') == 3)`, which turns up once `apply_variables` has substituted a saved rule's operands into literals) is correct Python and is left alone. The exemption keys on a recursive static-constant test — a literal, a literal collection, arithmetic over them, or a call to one of the sandbox's pure builtins whose args are themselves constant — so nothing that resolves to a column (always an `Attribute`/`Name`, never a constant) can ride in on it.

## A note for whoever answers the customer

`get_column(...) is None` is a Python identity test against a Column object: always `False`. That branch has never matched a row — the compiled SQL is literally `CASE WHEN false THEN ...`. Restoring `is` unblocks them; their null test still needs `.is_(None)`, and their output will change when they switch. That is the author's bug to fix, not the sandbox's to reject, so this PR does not try to be clever about it. A test pins the always-`False` behaviour so the guidance stays honest if SQLAlchemy ever changes.

## Verification

- `test_expression_sandbox.py` 30/30 pass, including every existing attack, the internals-shadowing case above, the no-`safe_dict` fallback, and the negated-collapse family — bare, compound `and`/`or`, pure-constant exemption, and a builtin-over-a-column that must stay rejected.
- `test_sql_expression.py` unchanged against baseline (88F/47P — pre-existing, environment-related, identical with and without this change).
- Both customer expressions compile to real SQL through `eval_expression` on a patched install.


simozzy:
## Self-review (adversarial pass on the full diff)

Automated claude-review is off on this repo, so this is my own review, run as an independent pass over both repos' diffs with instructions to attack the relaxation rather than approve it. Probes were executed against the patched code, not reasoned about on paper.

**Verdict: ship.** No sandbox escape found attributable to this change. `ast.Is` / `ast.IsNot` / `ast.Not` add no attribute, call or name-resolution surface; the underscore-column exemption never widens reachability beyond the specific `Column` object it resolves to.

Soundness of the identity check was confirmed at the SQLAlchemy level: `ColumnCollection.__getattr__` and `__getitem__` both resolve through the same `self._index[key][1]`, so for a non-colliding name the identity holds by construction, and for a colliding name (`_index`, `_all_columns`, `_colset`) Python's attribute protocol returns the real internal before `__getattr__` is ever reached — so the check correctly fails and the column stays blocked. Duplicate column names across a join (SQLAlchemy auto-suffixes) and `Column(key=…)` vs `name=…` divergence were both checked and behave correctly.

### Findings fixed before this note

1. **[medium] `_validate_ast` was not alias-scoped.** It took a flat set of underscore column names, so `table2._Flag` validated even when only `table1` owned `_Flag` — any name matching `_is_allowed_root` would do. Not exploitable, because both callers immediately re-run `_compile_ast`, which re-derives per alias from the real ctx and blocks it. Fixed anyway: the function documents itself as safe standalone and should not lean on a backstop a later refactor might remove. It now takes the same per-alias map the runtime guard uses. Verified in isolation: `table2._MajorAccountFlag` is rejected by `_validate_ast` alone.
2. **[low] CHANGELOG** in both repos covered only the `is`/`not` half.
3. **[low] No `eval_rule` coverage** for either relaxation — it is a separate call site that also had to be threaded with the eval context. Now pinned.
4. **[nit] Dunder filtering** lived in the caller on the utilities side and inside the helper on the plaid side. Aligned.

### Explicitly checked, clean

- Both `eval()` sites in `plaidcloud.utilities` (`eval_expression`, `eval_rule`) still call the guard with the eval context immediately before compiling. No new bypass.
- `plaid/app/ai/expression.py` calls no `eval()` at all — it is a hand-rolled AST interpreter — and fails closed on `is`/`is not`/`not` at the `_compile_ast` layer too, independently of the allowlist.
- No previously-working expression shape regresses (`case()`, `func.coalesce`, `.in_()`, `and_()`, arithmetic all still pass).

simozzy:
## Review loop complete — 3 adversarial rounds, verdict SHIP

Each round ran in fresh context with instructions to attack rather than approve, and probes were executed against the code, not reasoned about on paper. Rounds 1 and 2 each found a real defect; round 3 found only polish. Summary of what changed as a result:

### Round 1 — `not (a == b)` collapses silently

Restoring `ast.Not` brought in one shape that does **not** raise: SQLAlchemy's `BinaryExpression.__bool__` special-cases eq/ne and returns an `id()`-based Python bool. Verified live:

```
not (table.status == 'inactive')                        -> True   (plain bool)
case((not (table.status == 'inactive'), table.id), ...)  -> CASE WHEN true THEN t.id END
```

The column vanishes from the SQL and the branch becomes unconditional — silently wrong results in an allocation engine, which is worse than the loud failure every other `not` shape gives. Now rejected with a steer to `!=`.

### Round 2 — two more collapsing siblings

The round-1 fix caught only eq/ne. Both of these collapse the same way and were newly reachable for the same reason:

```
not (col is None)      -> Python identity, constant True
not (col in (1, 2, 3)) -> tuple.__contains__ bool()s the per-element `==`
```

The check now covers `==`, `!=`, `is`, `is not`, `in`, `not in`, and names the operator that actually builds SQL for each case (`!=`, `is_(None)`, `notin_([...])`, …) rather than one generic suggestion.

### Round 3 — hint selection on chained comparisons

A chained comparison expands to `(a op1 b) and (b op2 c)`, and `and` bool()s the left side — so **op1 alone** decides whether it collapses quietly or raises. Scanning every op meant `not (a < b == c)`, which raises loudly at the `<` and is therefore safe, got rejected as a silent collapse and handed an unrelated `!=` steer. Now reads `ops[0]`.

### Filed separately, deliberately NOT fixed here — sc-23189

The **bare** forms collapse too, and have shipped for years:

```
col in (1, 2, 3)              -> plain bool; CASE WHEN false THEN ...
(a == 1) and (b == 2)         -> only `a = 1` survives — second condition dropped
(a == 1) or (b == 2)          -> only `b = 2` survives — FIRST condition dropped
bool(a == 1)                  -> plain bool
```

`a == 1 and b == 2` is the most natural way anyone writes a two-condition filter, and it silently returns the wrong rows. Rejecting these is the right end state, but it fails saved expressions on upgrade, so it needs its own release note and — more importantly — a sweep of existing customer expressions to size the wrong-results blast radius first. That is sc-23189, and it is worth prioritising.

The distinction this branch holds to: `not` was blocked outright in the shipped release, so tightening the negated half regresses nobody. The bare half is a behaviour change and does not belong in a regression fix.

### Verified clean across all three rounds

No sandbox escape on any axis. The underscore-column identity check was re-attacked each round: a column named `_index` / `_all_columns` / `_colset` stays blocked (attribute access resolves to the internal, so the identity check correctly fails), while `_MajorAccountFlag` resolves to the real `Column`. Per-alias scoping holds. Both `eval()` sites still receive the guard with the same namespace the eval runs against. Every steer message's suggested replacement was compiled and confirmed to produce correct SQL. No stale callers of the renamed/re-signatured helpers anywhere in the tree.

Final: utilities sandbox 26/26; plaid AI suite 942; pylint clean at the pinned CI version; wider `test_sql_expression.py` unchanged at its pre-existing 88F/47P environment baseline.

simozzy:
## Production evidence: verified against a live customer victim (USESI, tenant-v5.11.5)

A read-only sweep of customer tenants turned up a real step this fix is needed for, and I ran the branch code against its actual expressions.

**The step:** "SOO Data Stage" (`zpukdqel`), a `frame_extract` that is **step 7 in an active, enabled workflow** in *USESI v03 - MASTER* (and its copy). It builds a sales/GP reporting table; 5 target columns sum account-code columns like `table._210 + table._215 + …`, where `_210`/`_215`/… are **real numeric columns** (account codes prefixed with `_` to be valid identifiers) — the customer's `_MajorAccountFlag` shape exactly. The tenant's deployed validator rejects them today: `disallowed_attr: Access to private attribute '_210' is not allowed`.

**Verified the fix covers it** (ran the branch, not reasoned):

- Runtime guard (this repo): `table._210 + …` compiles to real SQL for all five expressions.
- Validator (plaid `plaid/app/ai/expression.py`): `validate_expression` returns `ok: true` for all five.
- Adversarial hold: `table._index` (a name resolving to a `ColumnCollection` internal, not a column) still blocks — the identity-based exemption didn't over-open.

**Timing nuance (measured):** that step's last run *succeeded* (2026-07-16). So on USESI the validation surface already rejects it but the workflow runner hasn't broken yet — a latent victim this fix **prevents** breaking once the runtime hardening reaches the runner (it already fails at run time on the EMEA tenant that opened the investigation). Same fix, different timing per tenant.

Net: real customer data in a live reporting workflow confirms the underscore-column half of this change is necessary and correct, and that an internal-colliding name stays blocked.

plaidcloud/utilities/sql_expression.py
```diff
@@ -109,6 +118,51 @@ def _attr_off_func(node):
...
+    Only the FIRST operator matters: Python expands a chained comparison to
+    `(a op1 b) and (b op2 c)`, and `and` bool()s the left side, so op1 alone
+    decides whether the expression collapses quietly or raises. `not (a < b
+    == c)` raises at the `<` — loud and safe — and must not be mistaken for
+    the silent shape and given an unrelated `!=` steer.
+    """
+    if not (
+        isinstance(node, ast.UnaryOp)
+        and isinstance(node.op, ast.Not)
+        and isinstance(node.operand, ast.Compare)
```
rad-pat: **The silent-collapse guard misses the compound shape, and that shape is newly allowed by this PR.**

`isinstance(node.operand, ast.Compare)` only matches a bare comparison. Wrap the same comparisons in an `or`/`and` and the operand is a `BoolOp`, so the hint returns `None` and the expression sails through — but it collapses just as silently, because `and`/`or` call `bool()` on the operand and hand back a `BinaryExpression`, whose `__bool__` is the same hash-based special case.

Verified on this branch (SQLAlchemy 2.0.48):

```python
>>> se.eval_expression("not (table.a == 1 or table.b == 2)", {}, [t])
True                      # a plain Python bool, columns gone
>>> se.eval_expression("not (table.a == 1 and table.b == 2)", {}, [t])
True
>>> r = se.eval_expression("case((not (table.a == 1 or table.b == 2), 1), else_=0)", {}, [t])
>>> print(r.compile(compile_kwargs={'literal_binds': True}))
CASE WHEN true THEN 1 ELSE 0 END
```

That is precisely the artifact the comment below this function says the guard exists to prevent ("the branch becomes a constant (`CASE WHEN true THEN ...`)"), and on master it could not happen at all — `ast.Not` was not on the allowlist, so the whole family was rejected. Allowing `ast.Not` without covering the compound case turns a loud rejection into wrong SQL. `not (a == 1 or b == 2)` is, if anything, a likelier thing for an author to write than `not (a == 1)`.

Suggest recursing into the operand rather than matching only its top node: walk the operand subtree for any `Compare` whose first op is in `_NEGATED_COMPARISON_HINTS`, and when the operand is not a bare comparison steer to `~(...)` / `not_(...)`, which builds a real SQL NOT over the whole clause. A test alongside `test_negated_comparisons_are_rejected_not_silently_collapsed` covering `not (a == 1 or b == 2)` and `not (a == 1 and b == 2)` would pin it.

plaidcloud/utilities/sql_expression.py
```diff
@@ -109,6 +118,51 @@ def _attr_off_func(node):
...
+    ):
+        return None
+    return _NEGATED_COMPARISON_HINTS.get(type(node.operand.ops[0]))
```
rad-pat: **The rejection also fires on constant-vs-constant comparisons, which are correct Python and were fine before.**

The guard runs on the post-`apply_variables` source, so a saved rule whose operands are substituted literals now hits it:

```python
>>> se.eval_rule("not ('{v}' == 'CORP')", {'v': 'ABC'}, [t])
SQLExpressionError: `not (...)` around a comparison does not build a SQL NOT ... Use `a == b`.
>>> se.eval_rule("not ('{v}' in ('CORP','LLC'))", {'v': 'ABC'}, [t])
SQLExpressionError: ... Use `a.notin_([...])`.
>>> se.eval_expression("not (len('abc') == 3)", {}, [t])
SQLExpressionError: ... Use `a != b`.
```

None of these collapse anything — there is no Column involved, `bool()` is well-defined, and the expression means exactly what it says. `eval_rule` returning a plain Python bool is a supported path (`test_rule_path_gets_the_same_relaxations` asserts it), so this is a live shape. It is the same class of over-block this PR is fixing, and the `in` steer is actively wrong advice: a Python `str` has no `.notin_`.

Cheapest narrowing that keeps the real fix intact: skip the rejection when every operand of the comparison is a literal (`ast.Constant`, or a `List`/`Tuple` of them). `not ('ABC' == 'CORP')` and `not ('ABC' in ('CORP','LLC'))` both qualify; nothing touching a column does.

rad-pat (changes requested):
## Review

I read the diff, ran the suites, and attacked the new underscore exemption directly. **The two regression fixes are right and the security-sensitive part holds up. The extra `not (...)` guard added on top has a hole that lets silently-wrong SQL through — that's what I'd like fixed before merge.**

### What I checked and agree with

**`ast.Is` / `ast.IsNot` / `ast.Not` back on the allowlist** — correct. Pure operators, no attribute access, call, or name resolution; no escape surface. `ast.IfExp` staying out is the right line to hold.

**The underscore-column exemption keyed on resolved identity** — this is the part that could have re-opened sc-22664, and it doesn't. I built a hostile table with columns named after every `ColumnCollection` and `Column` internal I could find and checked what the guard lets through:

```
_index, _collection, _colset, _all_columns          BLOCKED   (resolve to the internal, not the column)
_data, _table, _proxies, _element, _type, ...       exempt -> sqlalchemy.sql.schema.Column
_sa_instance_state, _MajorAccountFlag               exempt -> sqlalchemy.sql.schema.Column
```

Every exempted name resolves to a `Column`; nothing hands back a `dict`, `list`, or collection internal. `getattr(collection, name, None) is collection[name]` is doing exactly the work claimed for it, and catching that the naive name-based version was exploitable — by attacking your own first cut — is the good call in this PR. Per-alias scoping, the dunder exclusion, and the closed-by-default `safe_dict=None` fallback are all covered by tests.

**Verification claims check out.** `test_expression_sandbox.py` 26 passed / 116 subtests. `test_sql_expression.py` is byte-identical between this branch and master in my environment (94F/49P here vs your 88F/47P — different env, same delta: zero).

**The changelog note about `is None` being a dead branch** is the right thing to have written down, and pinning it with a test so the guidance stays honest is a nice touch.

### What needs to change

**1. `not (a == 1 or b == 2)` slips past the new guard and compiles to `CASE WHEN true`.** (inline on `_negated_comparison_hint`)

The check matches only a bare `ast.Compare` operand. Put the same comparisons behind `or`/`and` and the operand is a `BoolOp`, so nothing fires — but `and`/`or` `bool()` their operand and return a `BinaryExpression`, so `not` collapses it just the same:

```python
>>> print(se.eval_expression("case((not (table.a == 1 or table.b == 2), 1), else_=0)", {}, [t])
...       .compile(compile_kwargs={'literal_binds': True}))
CASE WHEN true THEN 1 ELSE 0 END
```

This matters more than a missed edge case: on master `ast.Not` wasn't allowlisted, so this expression was *rejected*. After this PR it's accepted and produces wrong SQL. Allowing `ast.Not` without covering the compound shape trades a loud failure for a silent one, in the one area this PR went out of its way to make safe.

**2. The rejection fires on constant comparisons that are perfectly correct.** (inline on the hint lookup)

Because the guard runs post-`apply_variables`, `not ('{v}' == 'CORP')` with `v='ABC'` is now blocked, as is `not ('{v}' in ('CORP','LLC'))` — and that one steers to `.notin_([...])`, which a Python `str` doesn't have. Nothing collapses in either case; there's no Column involved. Same class of over-block this PR exists to fix.

### Nits, not blocking

- `_negated_comparison_hint` runs before the allowlist check in the walk, so `not (a == <disallowed node>)` reports the negation message rather than naming the actually-forbidden node. Cosmetic.
- The PR description doesn't mention the new `not (comparison)` rejection at all — it's a whole behavioural change reviewers would only find in the changelog or the diff. Worth a paragraph in the body.
- Small correction to the description's threat model, not to the code: `.table.metadata` is already reachable on master via `get_column(table, 'a').table.metadata` (non-underscore chain, unrestricted root), so `_all_columns[0].table.metadata` isn't new ground the exemption would have opened. The identity check's real value is blocking `_index` -> `dict` and `_all_columns` -> `list`, which is plenty of justification on its own.

Happy to re-review as soon as the compound-`not` case is covered.

rad-pat (changes requested):
See comments

simozzy:
## Both blockers fixed — pushed in `204bd4d`

Thanks for the attack — the compound hole was real and I've closed it. Everything ran through the code, not reasoned on paper.

### 1. Compound `not (a == 1 or b == 2)` — fixed

You were right: it compiled to `CASE WHEN true` and it was *newly* reachable because this PR restored `ast.Not`. The guard now walks the operand for any collapsing `Compare` instead of matching only a bare one, and a compound operand (comparisons under `and`/`or`) steers to `not_(and_(...))` / `not_(or_(...))`:

```
not (table.a == 1 or table.b == 2)   REJECT -> "... Use not_(and_(...)) / not_(or_(...)) ..."
not (table.a == 1 and table.b == 2)  REJECT
case((not (a==1 or b==2), 1), else_=0)  REJECT   # buried under a Call, walk still finds it
```

One deliberate detail: the steer is `not_(...)`, **not** `~(...)`. `~(a == 1 or b == 2)` doesn't help — the inner Python `or` collapses before `~` ever runs — so the message points at `not_()` over an explicit `and_()`/`or_()`, which I verified compiles to `NOT (a = 1 OR b = 2)`.

### 2. Constant-vs-constant over-block — fixed, and widened past the literal-only line

Exemption keys on a recursive **static-constant** test rather than name: a comparison is left alone when every operand is a literal, a literal collection, arithmetic over them, **or a call to one of the sandbox's 9 pure builtins whose args are themselves constant**. That last clause covers your `not (len('abc') == 3)` example too, not just the substituted-literal cases:

```
not ('ABC' == 'CORP')            ACCEPT -> True     (bare)
not ('ABC' in ('CORP','LLC'))    ACCEPT -> True     (your mis-steered `in` case)
not (len('abc') == 3)            ACCEPT -> False    (builtin over literals)
eval_rule("not ('{v}'=='CORP')", {v:'ABC'})  ACCEPT -> True
```

It can't widen into a hole because `_is_static_constant` never handles `Name`/`Attribute`/`Subscript` — a column reference is always one of those, so it bottoms out `False`. Anything touching a column stays rejected:

```
not (func.lower(table.a) == 'x')   REJECT   # your collapse case
not (sum([table.a]) == 0)          REJECT   # builtin *over a column* — arg isn't constant
```

I re-ran an adversarial pass over all 9 builtins applied to columns, nested/starred/BinOp wrapping, and buried BoolOps — every column-touching shape stays caught; no new escape surface (all those nodes were already allowlisted).

### Nits

- **`not (a == <disallowed node>)` reporting order** — left as-is. Fixing it cleanly means validating the operand subtree against the allowlist before the negation check, and `ast.walk` visits the `not` before its disallowed child, so it's not the one-liner it looks like. You flagged it cosmetic; happy to do it in a follow-up if you'd rather.
- **PR body / changelog** — updated the changelog bullet to describe the compound coverage and the constant exemption. Will add a paragraph to the body before marking ready.
- **`.table.metadata` threat-model correction** — agreed, it's already reachable on master via the non-underscore chain; the identity check's real value is `_index -> dict` / `_all_columns -> list`. Description only, no code change.

Verification: sandbox `test_expression_sandbox.py` 30/30; `test_sql_expression.py` unchanged at the pre-existing 88F/47P baseline (zero delta). Ready for another look.

rad-pat (changes requested):
Reviewed against PR head `204bd4d` — note the local `pr-335` checkout is behind (at `7e967ff`, missing the compound/constant-exemption commit). Ran the suite at head: **30 passed, 125 subtests**.

## Overview

Two false positives from the sc-22664 sandbox hardening:

1. `ast.Is` / `ast.IsNot` / `ast.Not` were omitted from `_ALLOWED_NODES` → restored, with `not (<collapsing comparison>)` newly rejected with a steer.
2. Underscore-led attribute access was blanket-blocked → now exempted per-alias, keyed on **resolved identity** (`getattr(collection, name, None) is collection[name]`), which requires threading `safe_dict` into `_assert_safe_expression`.

The identity-keyed exemption is the right design, and finding `_index` / `_all_columns[0].table.metadata` by attacking the naive name-based version is exactly the diligence this file needs. The write-up is unusually good.

## Security — no concerns

The relaxation is tight, and I couldn't widen it:

- `collection[name]` on a str key always returns a `Column`, so the exemption never exposes an object type that non-underscore columns didn't already.
- `_attr_is_column` requires `isinstance(node.value, ast.Name)`, so no chaining off the result.
- Dunders never exempt; `safe_dict=None` is closed-by-default; both real call sites (`eval_expression:310`, `eval_rule:1487`) pass the same dict `eval()` receives.
- `ReadOnlyColumnCollection` subclasses `sqlalchemy.sql.base.ColumnCollection`, so the `isinstance` filter catches `table.columns`, `tables_by_alias` reps, and the post-filter `extra_keys={'result': ...}` collection alike.

## Issues

### 1. `_negated_comparison_hint` walks the whole operand subtree — false positives with a wrong steer (blocking)

`ast.walk(node.operand)` flags *any* nested collapsing comparison, including ones consumed correctly as SQL. Confirmed at head:

```
not case((table.a == 1, true), else_=false)   BLOCKED, steer: "`a != b`"
```

That inner `table.a == 1` never collapses — `case()` takes the `BinaryExpression` intact. The expression raises `TypeError: Boolean value of this clause is not defined` at runtime, i.e. it belongs to the *loud and safe* family the code says it leaves alone (`sql_expression.py:231-233`). The author wrote a `case`, not a comparison to negate, and gets told to use `!=`. Same root cause makes a bare comparison pick up the compound steer:

```
not (table.a == (1 and 2))    steer: "`not_(and_(...))` / `not_(or_(...))`"   # BoolOp is in the comparator
```

Suggest descending only through what actually propagates the operand's boolness — nested `UnaryOp(Not)` and `BoolOp.values` — rather than `ast.walk`, stopping at `Call`/`Subscript` boundaries. That also fixes the arbitrary `collapsing[0]` pick (BFS order ≠ source order) and the misapplied compound test in one go.

### 2. `isnot` vs `is_not` inconsistency (low)

```python
ast.Is:    '`a.isnot(None)`',
ast.IsNot: '`a.is_(None)`',
```

On SQLAlchemy 2.0.48 `isnot` still works with no deprecation warning, but it's the legacy synonym — pairing it with the modern `is_` is inconsistent. Prefer `a.is_not(None)`. Both hints also hardcode `None`, which is wrong for `not (table.a is table.b)`.

### 3. The silent-collapse policy is asymmetric (low, design)

The stated rationale for rejecting `not (a == b)` is silent collapse — but `and`/`or` collapse just as silently and stay allowed. Confirmed:

```
'table.a == 1 and table.b == 2'           -> BinaryExpression   # first comparison silently dropped
'table.a is None and table.b is not None' -> False              # bare constant
```

The second is `IDENTITY_AND_NOT[4]`, added by this PR as a *passing* case. Scoping to what `ast.Not` reopens is defensible, but restoring `is`/`is not` makes the second shape newly reachable, so I'd say so explicitly in the comment at `sql_expression.py:228-233` rather than leave "every other `not` raises… which is safe" reading as a complete account.

### 4. `_is_static_constant` doesn't account for shadowed builtins (nit)

`node.func.id in _SAFE_BUILTINS` ignores that `safe_dict` globals resolve *before* `__builtins__` — `merge(alias_keys, default_keys, table_keys, extra_keys)` could bind `sum`/`len`/`str` to a `ColumnCollection` via an alias or extra key. Failure mode is a loud `TypeError` (a collection isn't callable), not a silent collapse, so it's theoretical — but `and node.func.id not in (safe_dict or {})` closes it for free.

### 5. `_underscore_columns_by_alias` runs unconditionally (nit)

It's built on every `eval_expression`/`eval_rule` call, before `ast.parse`, even when the source has no underscore attribute. Measured on a 300-column table: **49µs vs 11µs** per guard call. Negligible in absolute terms, but it's free to gate — build it lazily on first `ast.Attribute` hit, or short-circuit on `'._' not in source`.

## Tests

Strong — attack corpus, the internals-shadowing case, cross-alias, no-`safe_dict` fallback, the steers-actually-compile checks, and pinning the always-`False` `is None` behaviour so the customer guidance stays honest. Gaps:

- No case for issue #1 (a `not` over a comparison consumed by `case()`/`func.*`).
- The underscore map is exercised only through positional `tables`; the `tables_by_alias` and `extra_keys={'result': ...}` paths (`sql_expression.py:1135`, frame_join_multi) reach `_underscore_columns_by_alias` too and aren't covered.

## Style

Comment density is high but matches the surrounding sc-22664 block, and the CHANGELOG paragraphs match their neighbours. No convention issues.

---

**Requesting changes on #1 only** — the rest are polish and are fine to fold in or defer. Happy to re-approve as soon as the hint walk is narrowed.

simozzy:
## All five addressed — pushed in `d40bab9`

Thanks for the thorough pass. #1 fixed; #2–#5 folded in. Ran four independent adversarial rounds against the new detection (rounds 1–3 each found a real silent-collapse the first narrowing had reopened — see the design note below — round 4 clean). Everything below was executed, not reasoned on paper.

### 1 (blocking) — `_negated_comparison_hint` over-walked

Replaced `ast.walk` with a structured descent (`_boolness_bearing_comparisons`) that follows only what carries the operand's boolness to the outer `not`:

- a **`Compare` is a leaf** — recorded, not descended into, so a BoolOp/Call inside a *comparator* (`table.a == (1 and 2)`) no longer adds a comparison or flips it to compound;
- a **clause-building `Call`** (`case`/`func.*`/`.in_`) is a leaf — `not case((a==1,...))` is left to raise loudly instead of being mis-steered;
- `and`/`or`, a nested `not`, and index/concat of a literal collection are followed.

Your two confirmed cases now behave:

```
not case((table.a == 1, true), else_=false)   -> not guard-rejected; raises loudly at eval
not (table.a == (1 and 2))                     -> REJECT, bare `!=` steer (not compound)
```

**Design note (why the descent covers every builtin, not a hand-picked few).** My first narrowing was too aggressive and reopened silent collapses my own adversarial passes caught in sequence: `not (a==1 or b==2)` inside a builtin, `not (a==1, b==2)[0]` (indexing/concatenating a literal collection hands the clause back unchanged), and `not len([a==1])` / `not min([a==1])` / `not str(object=a==1)` (single-arg and keyword forms). Rather than special-case which calls are "safe leaves," the rule is now: **any pure-Python builtin call (`_SAFE_BUILTINS`) is transparent** — descend its args *and keywords* — and only genuine clause-builders are leaves. A comparison reached **through a wrapper** (a builtin call, an index) gets a **generic `not_(...)` / `~(...)` steer** rather than a per-operator one — which is also the clean root fix for the "author wrote a `case`, got told `!=`" class you flagged: a per-op steer is only given when the operand *is* a bare comparison.

### 2 — `isnot` → `is_not`

Done. `ast.Is` now steers to `a.is_not(None)`, matching the modern `is_`. Left the `None`-oriented wording: `not (a is b)` between two columns is degenerate Python identity (always `False`, same always-`False` note the changelog already makes for `is None`), so `is None`/`is not None` is the only real case worth steering.

### 3 — asymmetry comment

The comment at the raise site now says outright that bare `a == 1 and b == 2` / `in (...)` collapse silently too but are **deliberately out of scope here** (they shipped for years; `ast.Not` did not), and points at sc-23189 — so it no longer reads as "every other `not` is safe, full stop."

### 4 — shadowed builtins

`_is_static_constant` now takes `safe_dict` and treats a builtin name shadowed by an alias/extra_key (`sum`/`len`/… bound to a ColumnCollection) as non-constant — the descent uses the same check, so a shadowed `bool`/`str` is a leaf, not transparent. It also now recognises a `BoolOp` of constants (`(1 and 2)`) as constant, so a pure-constant comparison using `and`/`or` isn't over-rejected.

### 5 — lazy underscore map

`_underscore_columns_by_alias` is built on the first underscore-led attribute instead of on every guard call (lazy on first hit, so it's robust to whitespace rather than a `'._' in source` test).

### Tests

Added: comparison consumed by a clause-builder left to raise; comparator-BoolOp gets the bare steer; every wrapper collapse route rejected (index/concat, bool/str/len/min/max/sum, keyword-arg); shadowed-builtin call is a leaf; pure-constant BoolOp operand left alone; underscore column exempt via `tables_by_alias` and via `extra_keys={'result': ...}`.

### One residual, called out deliberately

`not str(table.a)` and `not str(table.a < 1)` still collapse — but that's `str()` stringifying *any* argument, which already happens on a bare `str(table.a)` with no `not` at all. It's a general value-vs-clause hazard of having `str` in the sandbox, not a negated-*comparison* collapse, so it's out of scope for this regression fix and belongs with the sc-23189 sweep (or dropping `str`/`bool` from the builtin set). Flagging it rather than pretending it's covered.

Verification: sandbox `test_expression_sandbox.py` 38/38; `test_sql_expression.py` unchanged at the pre-existing 88F/47P baseline (zero delta). Ready for another look.

rad-pat (approved):
Re-reviewed at head `d40bab9`. **The blocking issue is fixed — approving.**

I re-ran the narrowed walk against the cases I raised and then attacked it in the opposite direction, looking for collapses the narrowing might have reopened. Probes executed against the branch, not reasoned about on paper.

## Blocking issue #1 — resolved

`_boolness_bearing_comparisons` replacing `ast.walk` does exactly what was needed. Both confirmed cases behave, and I couldn't find a clause-consumed comparison it still mis-flags:

```
not case((table.a == 1, true), else_=false)             PASS guard -> raises loudly at eval
not case((table.a == 1, true), else_=false).label("x")  PASS
not func.coalesce(table.a == 1, 0)                      PASS
not and_(table.a == 1, table.b == 2)                    PASS
not or_(...) / not not_(...) / not func.lower(table.a).in_(["x"])   PASS
not (table.a < 1 == table.b)                            PASS   (chained, safe op first)
```

Making `Compare` a leaf also fixes the comparator case — `not (table.a == (1 and 2))` now gets the bare `!=` steer rather than the compound one — and the `collapsing[0]` pick is no longer BFS-order-arbitrary since a bare comparison is the only shape that gets a per-op steer.

Treating *every* `_SAFE_BUILTINS` call as transparent rather than hand-picking safe leaves is the right call, and the three collapses your own rounds caught (`(a==1, b==2)[0]`, `len([a==1])`, `str(object=a==1)`) are exactly why. The generic `not_(...)` / `~(...)` steer for wrapper-reached comparisons is a better answer than a per-op steer that would name an operator the author never wrote.

## Items 2–5 — all landed

`is_not` paired with `is_`; the asymmetry comment now names the bare `and`/`or` collapse as deliberately out of scope and points at sc-23189; `_is_static_constant` and the descent both check `safe_dict` shadowing; the underscore map is lazy on first underscore attribute.

## Security — re-attacked after the lazy-map change, still clean

Rebuilt the hostile table against a wider set of `ColumnCollection` internals and checked what each name resolves to:

```
_index -> dict, _collection -> list, _all_columns -> list, _colset -> set,
_parent -> DedupeColumnCollection, _proxy_index -> defaultdict,
_populate_separate_keys -> method                      ALL BLOCKED
_initiator -> Column, _Under -> Column                 exempt (real columns)
```

Every exempted name resolves to a `Column`; nothing hands back a collection internal or a bound method. Making the map lazy didn't change reachability — the exemption still works through positional `tables`, `tables_by_alias` and `extra_keys={'result': ...}`, dunders stay blocked, chained access off an exempted column stays blocked, and `safe_dict=None` is still closed-by-default.

## Verification

- `test_expression_sandbox.py` at head: **38 passed, 139 subtests**.
- `test_sql_expression.py`: 94F/49P on this branch and 94F/49P on `master` in my environment — **zero delta**, same pre-existing environment baseline.
- No callers of the re-signatured helpers outside `sql_expression.py`.
- Every steer the messages name compiles to a real clause (`!=`, `is_`, `is_not`, `in_`, `notin_`, `not_(and_/or_(...))`, `~(...)`).

## Non-blocking, for the sc-23189 sweep

**1. The wrapper collapse isn't operator-dependent, but the check is.** I swept 330 negated shapes over a column; 10 passed the guard and still returned a plain Python value. All are the same family, and all involve a *non-collapsing* operator under a scalarising wrapper:

```
not (str(table.a < 1))        -> False
not (len([table.a > 1]))      -> False
not (str(object=table.a < 1)) -> False
not (abs(len([table.a < 1]))) -> False
```

You disclosed `not str(table.a < 1)`, and your framing is right — `str()`/`len()` scalarise *anything*, so this is the value-vs-clause hazard of having them in the sandbox rather than a negated-comparison collapse. Worth noting the shape precisely for the sweep though: once a wrapper is in the chain the collapse no longer depends on `ops[0]`, so `_comparison_collapses` gating on the hint map is the reason these slip. If `str`/`bool` get dropped from `_SAFE_BUILTINS` under sc-23189 most of this family goes with them.

**2. `_is_static_constant` doesn't handle `ast.Subscript`, so a correct constant negation that indexes a literal is over-rejected:**

```
not ('{v}'[0] == 'A')            with v='ABC'   REJECT   (correct Python, no column)
not (('{v}', 'x')[0] == 'CORP')  with v='ABC'   REJECT
not ([1, 2][0] == 1)                            REJECT
```

Same over-block class this PR exists to fix, just a much narrower shape than the ones you already covered — it needs an index into a literal in a saved rule. A `Subscript` whose `value` and `slice` are both static constants would close it, and it can't widen anything since a column reference is still never a constant. Fine as a follow-up.

**3.** The negation check still runs before the allowlist check in the walk, so `not (a == <disallowed node>)` reports the negation message. Cosmetic, as before — happy to leave it.

Nice work on this one. The write-up on why the exemption keys on resolved identity rather than name, and the production evidence from USESI, are both above the bar for what this file usually gets.

PR #336, opened 2026-07-22
feat(connect): forward verify_ssl from PlaidConnection (sc-23168)

inviscid:
Ref: [sc-23168](https://app.shortcut.com/plaidcloud/story/23168)

## The gap

`plaidcloud-rpc` 1.12.0 made `verify_ssl` a real connection option — but `PlaidConnection` didn't forward it, so **the option was unreachable from the class UDF and notebook authors actually construct**. That is the same shape as the bug sc-21805 fixed: an argument accepted and silently discarded.

Without this, turning verification on for a Jupyter tenant means editing `__PLAID_VERIFY_SSL__` in tenant values — which is cp-rest-generated and gets overwritten — rather than opting in per connection.

## Why unconditional, not `rpc_uri`-gated

`project_id` and `workspace_uuid` are forwarded only alongside `rpc_uri`, because plaidcloud-rpc raises if they arrive without it. `verify_ssl` is different — `PlaidConfig` applies it *after* selecting a configuration source:

```python
if verify_ssl is not None:
    self.verify_ssl = bool(verify_ssl)
```

So it composes with the environment and `plaid.conf` paths too. It says how to *treat* the connection, not what the configuration *is* — same reasoning as `token_provider`.

## Tests

Two new cases, 12 passing. `False` is asserted explicitly: it is meaningful and must not be swallowed as falsy by a truthiness check.

Requires `plaidcloud-rpc >= 1.12.0`, which the current floor (`>=1.11.0`) doesn't guarantee — but every consumer bumping for this will take 1.12.0 anyway, and passing `verify_ssl` to an older rpc would be a `TypeError` rather than a silent no-op.


inviscid (commented):
✅ Reviewed directly — the change is one name added to a tuple, plus two tests and a CHANGELOG line.

Verified before merging:

- **`verify_ssl` belongs in the unconditional set, not the `rpc_uri`-gated one.** Checked against plaidcloud-rpc master: `PlaidConfig.__init__` applies it *after* the source dispatch — `if verify_ssl is not None: self.verify_ssl = bool(verify_ssl)` (config.py:460-461) — so it composes with the environment and `plaid.conf` paths. Gating it on `rpc_uri` like `project_id`/`workspace_uuid` would have made it unreachable for exactly the callers that need it (Jupyter, which configures from the environment).
- **`False` is not swallowed.** The forwarding uses `if name in kwargs`, i.e. presence rather than truthiness, so `verify_ssl=False` is passed through rather than dropped. Pinned by its own test.
- 12 tests pass; Ruff, ReviewDog/pytest, status and triage all green.

This closes a gap that would otherwise have forced tenant-values edits to enable verification — and those files are cp-rest-generated, so such edits get overwritten on the next republish.

PR #337, opened 2026-07-23
fix(sql_expression): treat a present-but-null config value as not supplied

inviscid:
`get_select_query`'s `fill_in` helper fell back to `config.get(var_name, default)`. `dict.get` only applies its default when the key is **absent** — a key present with a `None` value returns `None`. So a config carrying `limit_target_end: None` reached `lim = limit_target_end - off` and raised `TypeError`.

This is reachable from the product today. The Table Explorer's Max Rows control is a free-text combobox with no validator, so clearing it makes `parseInt` return `NaN`, `JSON.stringify` serialises that as `null`, and the server writes the key unconditionally. The result is an HTTP 500 on **Download → Limited Row Count** and on **Save as View**, and a misleading "maybe something's wrong with your advanced filter" alert on grid refresh.

## The fix

```python
if var is None:
    var = config.get(var_name)
return default if var is None else var
```

The helper's first line already treated an explicit `None` *argument* as "not supplied"; only the config path disagreed. This makes the two consistent — a repair, not a new semantic.

It guards nine parameters, and seven are provably unaffected: `aggregate`, `having`, `use_target_slicer`, `distinct`, `count` and `disable_variables` are each consumed by a truthiness check, and `aggregation_type` is unaffected because its signature default is `'group'` rather than `None`, so its config lookup is reached only when a caller passes the argument as `None` explicitly. Only the two limit parameters change behaviour, which is the bug. Falsy-but-not-None values are untouched: `0` still means `0`.

It also fixes an unclaimed sibling case — `limit_target_start: None` previously raised the same `TypeError` and now yields `OFFSET 0`, which rescues three call sites in `portable_export_capture.py` that use `.get(k, 0)`.

## Testing

Three new tests, each verified to fail without the fix (or, for the neutrality claim, to fail under a mutation of the default). The all-nine neutrality test forces `use_target_slicer=True` as an argument so the limit block actually executes — an earlier version passed vacuously because a null slicer skipped the block entirely.

Full suite: 701 passed, 10 skipped.

## Shipping note

The consumer needs a matching clamp before this reaches users. On its own, a null limit now compiles to `LIMIT 0`, and Save as View **persists** the compiled SQL — so a transient typo would become a permanently empty view. The paired plaid PR clamps a missing row count to the explorer's existing 5000 default. Cut `v1.13.2` from this, then land the pin bump and the clamp together.


inviscid (commented):

Independent clean-room review. Traced all nine parameters the helper guards — including through `get_from_clause`, `eval_expression`, `constant_from_clause`, `expression_from_clause` and `get_combined_wheres` — and confirmed the seven non-limit params are consumed only via truthiness checks (or, for `having`, a default that is already `None`), so `None` versus the default is behaviourally identical everywhere in this repo and at every `plaid` call site. Falsy-but-not-None values are preserved: `0` still means `0`.

Verified red-without-fix by reverting only the change in a disposable scratch copy (via `git checkout <sha> -- <path>`, never `git stash` — `refs/stash` is shared across worktrees): both limit tests fail with the original `TypeError`, and the neutrality test fails under a mutation of the `aggregation_type` default. Confirmed `fill_in` is defined once and that the sibling `modified_select_query` does not carry the same trap — it runs `filter_nulls` first, which is the same present-null-means-absent convention this fix adopts. Full suite 701 passed, 10 skipped.

One pre-existing observation, out of scope and unchanged by this PR: `aggregation_type`'s config lookup appears unreachable from any real caller, since its signature default is `'group'` rather than `None` — so config-driven `rollup`/`sets`/`cube` selection has apparently been dead. Ticketed separately.

PR #338, opened 2026-07-23
fix(sql_expression): don't emit an empty ROLLUP()/CUBE()/GROUPING SETS()

inviscid:
`GROUP BY ROLLUP()` is not valid SQL. Neither is `CUBE()` or `GROUPING SETS()` — the empty grouping set is spelled `GROUP BY ()`. All three are rejected by the Postgres grammar (verified with `pglast`/libpg_query, not by eye), and Databend and StarRocks reject them too.

`get_select_query` emitted exactly that whenever a special aggregation type was selected and **no output column was marked as a grouping column**. Plain grouping degraded harmlessly in the same case — `group_by(*[])` renders no `GROUP BY` at all — so only the three special forms were affected, and only on that one shape.

## This is broken in production now

22 call sites forward `aggregation_type` today, 21 of them with a user-controlled aggregation (the Sage export hardcodes plain grouping, so it can't hit this): the inner/outer join family, union, union-all, upsert, loader, lookup, both allocation steps, postman, remote-import-table, rule-based tagging, condition checks, forecast, and the Excel/Avro/Parquet/Templated/XML exports. Each of the 21 takes user-controlled `target_columns` alongside the forwarded aggregation type — `frame_join_inner.py:52-55` and `export_excel.py:35-39` are representative.

So a step with Summarize on, Rollup (or Grouping Sets, or Cube) selected, and every output column left as a real aggregate fails against the currently pinned `plaidcloud-utilities==1.13.2`. It is not a latent trap.

## The fix

Guard the three special branches on non-empty `grouping_columns` and fall through to the plain branch:

```python
if not grouping_columns:
    select_query = select_query.group_by(*grouping_columns)
elif aggregation_type == 'rollup':
    ...
```

The fallback is character-for-character the statement the `else` branch already ran, so the no-grouping path is provably unchanged rather than merely believed to be.

## Evidence it changes nothing else

A 1536-case matrix (6 target-column shapes × `aggregate` ∈ {True, False, None, absent} × config type × argument type, both over {absent, None, group, rollup, sets, cube, '', bogus}) compiled against both this branch and master:

- **768 non-empty-grouping cases: 100% identical**
- 54 cases differ, and every one satisfies all four invariants — `aggregate=True`, a zero-grouping shape, master emitted an empty special form, this branch emits no `GROUP BY`

The 27-per-shape count decomposes exactly as 24 explicit-argument combinations plus 3 `argument=None` config-lookup combinations, which independently re-confirms the parameter-precedence semantics are untouched.

The new test fails without the guard for all three forms (`'GROUP BY' unexpectedly found in '... GROUP BY CUBE()'`).

## Docstring

`aggregation_type`'s documented values omitted `'cube'`, which the code has handled since it was added. It now also records the behaviour that made this easy to get wrong: unlike the other config-backed parameters, its non-`None` signature default means an omitted argument does *not* fall through to `config['aggregation_type']` — callers wanting the step config honoured must pass it explicitly. That is why 21 call sites silently discard a user's selection; fixing those is separate work.


inviscid (commented):

Independent clean-room review confirmed the bug against a real Postgres parser (pglast/libpg_query): `GROUP BY ROLLUP()`, `CUBE()` and `GROUPING SETS()` are all syntax errors, while `GROUP BY ()` and `GROUPING SETS(())` are valid — so all three special forms were affected, not two. The guard's non-empty path was proven inert: the fallback statement is character-identical to the `else` branch master already ran, and a compiled-SQL diff across representative cases showed zero differences on every non-empty-grouping case. The new test fails for all three forms without the guard (verified by reverting only the source in an isolated copy). The reachability is real and current: 22 call sites forward `aggregation_type` on today's pinned build, 21 with a user-controlled aggregation, so this is a live production defect rather than a latent one. Full local CI (liccheck, ruff, pytest --doctest-modules) is clean with no new findings.

One deliberate design choice, flagged not blocked: when a user selects Rollup with no grouping column, the fix degrades to a single ungrouped aggregate row rather than raising — consistent with what plain Group already does in that case, and correctly scoped for a low-level SQL builder. A step-level "this configuration produces no subtotals" warning belongs in the native qooxdoo form layer, not here, and is a reasonable follow-up.

PR #339, opened 2026-07-23
StarRocks forms for six Alteryx-converter functions (sc-23229)

inviscid:
## Why

The Alteryx expression converter passes unmapped functions through as `func.<name>`, emitting `array_tail`, `array_to_string`, `string_agg`, `titlecase`, `median`, and `any`. StarRocks has no equivalent for any of them, so SQLAlchemy rendered them **literally** and StarRocks rejected the query (`No matching function with signature: array_tail(...)`). This surfaced running the Alteryx corpus against a StarRocks workspace.

## What

| emitted | StarRocks | Databend / default |
|---|---|---|
| `array_tail` | `array_slice` | `slice` (Databend's real function) |
| `array_to_string` | `array_join` | native |
| `string_agg` | `group_concat(expr SEPARATOR sep)` | native |
| `titlecase` | `initcap` | **CompileError** — Databend has no title-case function |
| `median` | `percentile_approx(col,0.5)` | native |
| `any` | `any_value` | native |

Five of the six have a genuine native Databend spelling, so their default rendering is left as the built-in. **`titlecase` is the exception**: Databend has only `upper`/`lower` (no `initcap`, no per-word title-case — confirmed against Databend's string-functions docs), so its default path would emit a literal `titlecase(...)` that fails at run time. Rather than that, or a silently-wrong first-letter-only approximation, it raises a clear `CompileError` on Databend (matching the file's existing `st_area`/`QUANTILE` patterns). StarRocks keeps its native `initcap`.

Worth a reviewer's eye:
- **`string_agg`** takes the delimiter as a `SEPARATOR` clause, not a second argument — passing it as an argument makes StarRocks concatenate it onto *every* value. Verified live.
- **`median`** → `percentile_approx` is approximate, matching Alteryx's own median on large inputs.

The rest of the converter's function set (`modulo`, `ord`, `today`, date extractors, the `add_*` family, `date_diff`, `to_string`, `regexp_substr`) is already handled on master, so only these six are added — no duplicate `GenericFunction` registration.

## Testing

`TestAlteryxDialectAdditions` (StarRocks) asserts each renders to its StarRocks form, including `string_agg` with and without a separator. `TestAlteryxDialectAdditionsDatabend` asserts `titlecase` raises `CompileError` on Databend and `array_tail` renders `slice` there. All six also verified against live StarRocks on paul-dev. Full `test_sqlalchemy_functions.py` failure set is byte-identical to master's baseline.

## Downstream

Release blocker for **PlaidCloud/plaid#6686**. Once this merges and a release is tagged, plaid's `plaidcloud-utilities` pin is bumped before #6686 can merge safely.


inviscid (commented):

Independent re-review confirmed the titlecase Databend fix is correct (matches the file's `st_area`/`quantile` CompileError pattern; StarRocks `initcap` intact; no import/other-dialect impact) and verified the other five additions are genuine native Databend functions (no hidden second gap). Mechanically scanned SQLAlchemy core + the pg/databend/starrocks/mysql/sqlite/mssql/oracle dialect packages for all six names — no duplicate registration; `array_tail` correctly avoids clobbering PostgreSQL hstore `slice`. Tests are real (no skips, real dialects) and assert the CompileError.

Post-approval hardening (the reviewer's own suggested one-liner): changed the Databend-only `@compiles(titlecase, 'databend')` to a **bare default** `@compiles(titlecase)` so StarRocks' `initcap` is the only specialization and every other dialect (greenplum/snowflake/...) fails loud rather than silently reintroducing the literal render. Tests cover Databend and a raw postgresql dialect both raising. Regression set unchanged from master baseline; zero duplicate-registration warnings.

PR #341, opened 2026-07-25
Render string CASTs as STRING on StarRocks, not CHAR (sc-23267)

inviscid:
[sc-23267](https://app.shortcut.com/plaidcloud/story/23267). Fixes converted Alteryx **Union** workflows on StarRocks.

## The bug

Converted Alteryx unions fail on StarRocks with:
```
(1064, 'Incorrect logical plan found in operator: LOGICAL_UNION ... input cols type not equal with output cols type')
```

Root cause, traced end to end:
1. `frame_union_all` casts every branch column to its dtype (`String` → `PlaidUnicode(4000)`).
2. `PlaidUnicode` falls through to `NVARCHAR` on StarRocks (no dedicated branch in plaidcloud-rpc).
3. The StarRocks dialect inherits **MySQL's CAST rules** — MySQL CAST forbids `VARCHAR`/`TEXT`/`STRING`, so every string cast renders as **`CHAR`**. `String(4000)` → `CAST(x AS CHAR(4000))`.
4. StarRocks `CHAR` maxes at **255**, so `CHAR(4000)` is invalid — *tolerated* in a bare `SELECT` (coerced), but the strict `INSERT … UNION` type check rejects it.
5. The physical string columns are **`STRING`** (`varchar(1073741824)`). Union output `CHAR` ≠ target `STRING` → the error.

A bare `SELECT CAST(… AS CHAR(4000)) UNION …` runs fine live; only the INSERT-into-STRING path trips.

## The fix

```python
@compiles(sqlalchemy.sql.elements.Cast, 'starrocks')
def compile_cast_starrocks(element, compiler, **kw):
    if _cast_targets_string(element.type):   # unwraps PlaidUnicode TypeDecorator
        return f'CAST({compiler.process(element.clause, **kw)} AS STRING)'
    return compiler.visit_cast(element, **kw)
```

Renders a string-target CAST as `STRING` (matching the physical columns); non-string casts are untouched. Verified:
```
String cast  -> CAST(x AS STRING)          (was CHAR(4000))
NULL   cast  -> CAST(NULL AS STRING)
Int    cast  -> CAST(n AS SIGNED INTEGER)  (unchanged)
Decimal cast -> CAST(d AS DECIMAL(38, 10)) (unchanged)
```
`CAST(… AS STRING)` and its union are confirmed working live on StarRocks, and STRING matches the physical column, so the union INSERT type-checks.

## Blast radius

**StarRocks dialect only.** Databend/default/Snowflake CHAR rendering is unaffected (the CHAR-asserting tests are all on those dialects). Full utilities suite: **identical failure set with and without this change** (10 pre-existing environmental failures, 668 passing) — **zero new failures**.

## Tests

Added to `TestAlteryxDialectAdditions` (StarRocks): string cast → STRING, NULL → STRING, non-string casts unchanged; plus a Databend test asserting the rewrite does NOT apply there.

## Follow-on (separate)

Cut a utilities release and bump `plaidcloud-utilities` in `plaid` + `workflow-runner`, then verify Union.yxmd passes on paul-dev.


inviscid (commented):

Independent reviewer cloned PR #341 at head, ran the full suite live (214/214 pass), and empirically probed the StarRocks CAST override against the compiled dialect. Confirmed: the `@compiles(Cast, 'starrocks')` override is dialect-scoped and provably inert for databend/postgres/snowflake/greenplum; the length>255 boundary is correct at the edges (PlaidUnicode(255)→CHAR(255) unaltered, (256)→STRING, s8/cidr→CHAR unaltered, bare String/Text→STRING, GUIDHyphens→CHAR(36) excluded by isinstance); every changed test expectation is a clean CHAR→STRING substitution; and the databend/snowflake import_col golden masters in the same dict are byte-identical to master. Owner condition (StarRocks behavior may change, zero non-StarRocks impact) is satisfied.

PR #343, opened 2026-07-25
frame_join_multi: accept PlaidCloud dtype aliases + real column names (sc-23307)

inviscid:
Converted-Alteryx **multi-table joins** are rejected by `frame_join_multi`'s config validator before they can run, with `source_column_dtype_invalid` / `source_column_invalid`.

Two over-strict rules, both stricter than what the executor actually accepts:

1. **dtype** was checked against a small closed lowercase enum (`text`, `integer`, …). The converter and MCP/hand-authored steps legitimately emit PlaidCloud/pandas aliases — `String`, `Int32`, `Float64`, `datetime64`, `varchar`. The executor builds each column's type via `sqlalchemy_from_dtype` (regex-based, accepts all of these). The validator now validates dtype **by resolving through that same function** — drift-free — with the enum kept only as a fast path. Unresolvable dtypes (`bogus`, `''`) are still rejected.
2. **column id / target name** had to be a bare SQL identifier, so real Alteryx names with spaces/punctuation (`Position Title`, `Years (starting at 1st grade)`) were rejected. `get_table_rep` wraps each name in a `sqlalchemy.Column`, so SQLAlchemy's IdentifierPreparer quotes it at emit — spaces/punctuation are safe. The check is now a **break-out gate**: non-empty, ≤255 chars, and free of control chars and the quote characters every dialect uses (`\"`, backtick, backslash).

Adds `TestRealColumnNamesAndDtypeAliases` (spaced/punctuated ids accepted; quote/control/empty/overlong rejected; PlaidCloud dtype aliases accepted; unresolvable dtypes still rejected). Full validator suite green (62 passed, 66 subtests).

Ships via the utilities release + pin-bump chain; live end-to-end verification (converting + running Join_Multiple on a StarRocks workspace) will follow the deploy.


inviscid (commented):

Independent reviewer traced every emit path for source_columns[].id / target_columns[].target — all route through get_table_rep's `sqlalchemy.Column` / get_insert_query's `from_select` / `.label()`, and independently COMPILED adversarial names (quotes, backtick, backslash, `; DROP`) against both postgresql (Databend-like) and mysql (StarRocks-like) dialects: SQLAlchemy's IdentifierPreparer escaped every case. No raw string-format / f-string / apply_variables ever touches column names (apply_variables only applies to expression/constant/rule). The join-condition path stays gated by the untouched stricter _COLUMN_REF_RE. dtype fallback resolves via sqlalchemy_from_dtype and fails closed on any error (plaidcloud-rpc is a hard dep, so the lazy import can't legitimately ImportError). Tests cover the boundary.

Non-blocking follow-up (noted): add a compile-level test asserting get_table_rep quotes a punctuated id/target end-to-end, so the safety property is self-verifying rather than review-verified.

PR #346, opened 2026-07-26
sql_expression: fail closed on None-valued target-column expression (Alteryx is_sequence crash)

inviscid:
## Problem

A calculated target-column expression that compiles to Python `None` — an unmapped/empty Alteryx-converted formula, or a bare `None`/`null`/`NULL`/`Null` (all of which `get_safe_dict` maps to Python `None`) — was treated as valid. The `None` then failed in a way whose exact shape depended on dtype and SQLAlchemy version: a silent, wrong-data `CAST(NULL)` when a cast_type was present, or an `AttributeError` deep in SQLAlchemy's `INSERT … from_select` column scan (`'NoneType' object has no attribute 'is_sequence'`, or `… has no attribute 'label'` for `serial`/`bigserial` columns, whose cast_type is falsy) — in every case naming neither the step nor the column.

## Fix

`expression_from_clause` now fails **closed**: immediately after `eval_expression`, if the result is `None`, it raises a clear `SQLExpressionError` naming the target column and the source expression, before the value reaches `process_fn` or the INSERT scan. `SQLExpressionError` is the module's own type, already converted to a `UserError` by the workflow runner's `transform_handler`.

**Raise, not null-coerce** — rationale: bare `None` is not a supported "emit NULL" convention anywhere in this codebase (no test relies on it). A genuine NULL column is written as a real clause — `cast(null, <type>)` or `sqlalchemy.null()` — both of which evaluate to a `ColumnElement`, not bare `None`, and are unaffected by the guard. The legitimate `count_null` aggregate (source column set to `None`, wrapped in `func.count`) goes through `source_from_clause`, not this path. So the only thing the guard catches is the genuinely degenerate case.

## Tests

- `TestGetFromClause::test_expression_none_raises` — `expression` ∈ {`None`,`Null`,`null`,`NULL`} × `dtype` ∈ {`text`,`numeric`,`serial`,`bigserial`} → raises `SQLExpressionError` mentioning the column. The `serial`/`bigserial` cases specifically exercise the falsy-cast_type `.label` `AttributeError` path.
- `TestGetFromClause::test_expression_explicit_null_still_compiles` — `cast(null, integer)` and `sqlalchemy.null()` still compile to a `Label` (legitimate NULL unaffected).
- `TestGetSelectQuery::test_none_expression_column_fails_closed_at_select_build` — the full select→insert pipeline for a `None`-valued expression column raises `SQLExpressionError` at the select step, so the `None` never reaches `get_insert_query`.
- `TestGetSelectQuery::test_valid_expression_with_record_id_compiles_through_insert` — regression: a normal valid expression plus a RecordID `row_number().over()` serial compiles clean all the way through `insert().from_select(...)`, genuinely exercising the INSERT column scan that used to crash on a `None` element.

Full `test_sql_expression.py` suite: **151 passed, 32 subtests passed**, matching a clean origin/master baseline (0 failures) in the same environment. No new ruff findings vs master.

## Known separate degeneracy (not in scope)

A target column with a missing/`None` `dtype` still raises upstream in `get_from_clause` at `sqlalchemy_from_dtype` (`RegexMapKeyError: 'none'`) — before `expression_from_clause` is reached — so this guard's location structurally cannot intercept it. That is a distinct invalid-dtype bug, not the expression→None path fixed here.

## Release

Version is managed by `setuptools_scm` (git tags); no version file to bump. Latest tag `v1.19.0`; CHANGELOG entry added under `## Unreleased`. Consumed by plaid-rpc + the workflow executor image via a version bump — release/downstream pin to be coordinated separately.


inviscid (commented):
✅ Independent Sonnet review (COMMENT, no blockers) + accuracy items addressed — **APPROVE**

The guard is verified correct: it fires immediately after eval_expression and strictly before process_fn/select()/insert().from_select(), so it can't miss the None propagation paths; legitimate NULL columns (cast(null,type)/sqlalchemy.null()) evaluate to real ColumnElements and pass untouched; count_null bypasses this path entirely (source_from_clause); and the dtype-None degeneracy is structurally upstream (RegexMapKeyError before this branch) — correctly scoped out (handled at convert time by the plaid Fix B). All 3 review nits fixed: test renamed to honestly assert select-build-time + genuine INSERT-compile coverage added; serial/bigserial+None regression added (reproduces the distinct `.label` pre-fix crash); wording softened to not overclaim the exact is_sequence signature. 151 passed, 32 subtests, matches origin/master baseline.

PR #347, opened 2026-07-27
Drop unused runtime deps: argparse (stdlib), setuptools, avro

inviscid:
Three hard runtime dependencies that nothing in this package uses:

- **`argparse`** — in the Python standard library since 2.7. Pinning it installs a redundant PyPI backport into every consumer. `clean_files.py` imports argparse, which resolves from stdlib either way.
- **`setuptools`** — zero runtime imports of `setuptools` or `pkg_resources` anywhere in `plaidcloud/` (verified repo-wide).
- **`avro`** — no module imports it. `frame_manager.avro_from_sql` only maps SQL types to Avro type *name strings*; it never touches the library. (Consumers that genuinely need Avro get `fastavro` via the `[full]` extra.)

Verified: fresh venv install from the trimmed requirements, then imported frame_manager / sqlalchemy_functions / data_helpers / stringtransforms / clean_files — all OK, and `avro_from_sql` still works with the `avro` package absent.

Part of the rpc image-slimming effort — this is the shared-library half.


inviscid (commented):

Verified each of the three claims directly against the plaid-utilities source rather than trusting the PR description, and cross-checked consumer repos in the PlaidCloud org for transitive-dependency risk.

- `argparse`: only importer is `plaidcloud/utilities/clean_files.py:7` — pure stdlib usage, and `pyproject.toml` sets `requires-python = ">=3.10"`, far above the 2.7 threshold where argparse entered the stdlib. Correct removal.
- `setuptools`: no `import setuptools` / `pkg_resources` anywhere under `plaidcloud/` at runtime — the only hit is `setup.py`'s build-time import, already declared in `[build-system] requires`. Consumers that need `pkg_resources` (plaid, workflow-runner, RediStomp) pin setuptools themselves. Correct removal.
- `avro`: no `import avro` / `from avro` anywhere in the package. `frame_manager.avro_from_sql` (:457-480) is a dict lookup mapping SQL type strings to Avro type-*name* strings; it never touches the library. `PlaidCloud/plaid` does real avro work but pins `avro==1.12.1` directly, so it doesn't rely on this package to supply it.

No other consumer repo imports `avro`, `setuptools`, or `pkg_resources` without its own explicit pin, so none would break at import time from this trim. No findings block merge.

PR #350, opened 2026-07-27
sql_expression: coerce None-valued target-column expression to SQL NULL instead of raising (fixes v1.20.0 regression)

inviscid:
## What this fixes

This fixes a **v1.20.0 regression** (PR #346) that turned previously-passing Alteryx-converted workflows into hard failures.

#346 added a guard to `expression_from_clause`: if a calculated target column's expression evaluated to Python `None`, raise `SQLExpressionError`. The intent was sound — stop a bare `None` from reaching SQLAlchemy and surfacing as an opaque `AttributeError` (`'NoneType' object has no attribute 'is_sequence'` / `.label`) or as a silent, ambiguous `CAST(NULL)`.

It over-corrected. On a live corpus (StarRocks 4.1.3, utilities 1.20.0), **8 workflows now fail** with:

```
Target column '<name>': expression 'None' compiled to None/NULL; a calculated column must produce a SQL clause
```

Affected, including two that were **passing before the guard**:

| Workflow | Status before v1.20.0 |
| --- | --- |
| `Union.yxmd` | **passing** |
| `Poly-Split Build.yxmd` | **passing** |
| `Join_Multiple` | failing differently |
| `Blob_Convert` | failing differently |
| `Download_and_parse_web_data` | failing differently |
| `Merge_to_a_master_file_with_fuzzy_matching` | failing differently |
| `Optimize_location 08` / `09` | failing differently |

## Why coercion, not raising

For a legitimate Alteryx `Null()`, an empty formula, or an unmapped-but-harmless column, **SQL NULL is the correct output**. The pre-v1.20.0 behavior (`CAST(NULL AS <type>)`) was right for all of those.

Critically, **the executor cannot distinguish "intentional NULL" from "unmappable formula"** — by the time `expression_from_clause` sees it, both are just Python `None`. Raising therefore fails the intentional case to catch the unintentional one.

The fail-closed signal for genuinely unmappable formulas **already exists upstream, at convert time**, in `plaid`: `apply_formulas_to_target_columns` (`plaid/app/analyze/utility/alteryx_expression_converter.py`) already detects a clause-less conversion, emits an explicit `sqlalchemy.null()` so the column still exists (dropping it would leave a dangling downstream reference), collects the raw formula text into `unmapped_formulas` for a `runtime_unmapped_note`, and lowers the step's confidence. That is the correct enforcement point — it still has the Alteryx source text and can tell the two cases apart. The executor does not need to be a second enforcement point, and cannot be a correct one.

## What the guard does now

`expression_from_clause` **coerces instead of raising**: when `eval_expression` yields Python `None`, it substitutes `sqlalchemy.null()` and emits a `WARNING` naming the target column and the source expression. `process_fn` then casts that NULL to the column's dtype exactly as it did before v1.20.0, so the emitted SQL is byte-for-byte the pre-regression clause:

```sql
CAST(NULL AS NUMERIC) AS "Calc"
```

**Both reasons the v1.20.0 guard existed are preserved:**

1. The bare `None` still never reaches SQLAlchemy's `INSERT … from_select` column scan — no opaque `AttributeError`. (`sqlalchemy.null().label(name)` is also safe on the no-`cast_type` path, where bare `None.label` used to raise.)
2. The NULL is no longer silent or ambiguous — it is logged, with the column name and the source expression.

**Unchanged:** a missing or unresolvable `dtype` still raises upstream in `get_from_clause` at `sqlalchemy_from_dtype` — that is a separate degeneracy with a clearer error, and this PR does not touch it. The `count_null` path goes through `source_from_clause`, not this guard, and is likewise untouched.

## Tests

The #346 tests asserted `SQLExpressionError` for `'None'`/`'Null'`/`'null'`/`'NULL'` across `text`/`numeric`/`serial`/`bigserial`. Those assertions are now wrong, so they are **inverted**:

- `test_expression_none_coerces_to_sql_null` — each expression × dtype now compiles to a labelled SQL NULL clause, **and** a warning is logged naming the column and the expression (matched on the reprs, so the message boilerplate can't satisfy the assertion).
- `test_expression_none_casts_to_column_dtype` — the coerced NULL is cast to the column's dtype, byte-identical to a hand-written `cast(null, <type>)`.
- `test_expression_explicit_null_still_compiles` — an explicit `cast(null, integer)` / `sqlalchemy.null()` takes the normal path, with `assertNoLogs` proving no coercion happened.
- `test_expression_valid_is_unchanged` — regression: a valid expression and `func.row_number().over()` still compile unchanged.
- `test_none_expression_column_compiles_to_null_through_insert` — the original `is_sequence` reproduction, now asserting the **coerced** outcome: full `get_select_query` + `get_insert_query(...).from_select(...)` compiles to valid SQL containing `CAST(NULL AS NUMERIC) AS "Calc"`, with no raise and no `AttributeError`.
- Kept `test_valid_expression_with_record_id_compiles_through_insert` from #346 unchanged.

### Results

- `plaidcloud/utilities/tests/test_sql_expression.py`: **153 passed, 0 failed** (baseline on `origin/master` was 151 passed; net +2 tests).
- Full `plaidcloud/utilities/tests/`: **559 passed**, with 25 pre-existing failures in `test_sqlalchemy_functions.py` (snowflake dialect only). Those 25 reproduce identically on clean `origin/master` (557 passed / 25 failed) and are **unrelated to this change** — excluded.
- `ruff check` on both changed files reports the same 7 + 1 pre-existing findings as `origin/master`; **zero new** findings.
- Mutation-checked: reverting `sql_expression.py` to `origin/master` while keeping the new tests fails 20 of them, confirming the tests genuinely exercise the change.

## Release

Not tagged or published here. Per this repo's convention (setuptools-scm; one minor bump per release — `v1.19.0`, `v1.20.0`, `v1.21.0` are consecutive single-PR tags), this should ship as the next minor tag. `v1.20.1` and `v1.21.0` are both unavailable — `v1.21.0` is already taken by #345 — and current `master` HEAD (#347) is itself untagged, so the release carrying this fix is **`v1.22.0`**.


inviscid (commented):
✅ Regression fix verified — **APPROVE**

Restores pre-v1.20.0 behavior exactly: the guard now substitutes `sqlalchemy.null()` + logs a WARNING instead of raising, and `process_fn` casts it to the column dtype so the emitted SQL is byte-identical to pre-v1.20.0 (`CAST(NULL AS NUMERIC) AS "Calc"`). Both original goals are preserved — the bare None never reaches SQLAlchemy (no opaque is_sequence/.label AttributeError) and the NULL is no longer silent (logged with column + expression). The fail-closed signal for genuinely unmappable formulas correctly stays at CONVERT time in plaid (apply_formulas_to_target_columns emits null() + runtime_unmapped_note), which is the only layer that can distinguish intent. Tests 153 passed (+2 vs master), mutation-checked (reverting the fix fails 20), 25 snowflake-dialect failures confirmed pre-existing. Also corrected two inaccurate claims carried over from #346.

PR #351, opened 2026-07-27
StarRocks: native mode (and median) instead of failing closed

inviscid:
## Problem

Alteryx **Summarize → Mode** fails closed on StarRocks:

```
mode (statistical mode aggregate) has no StarRocks equivalent;
run this workflow on a Databend workspace, or replace the Mode aggregation.
```

That is what the live corpus workflow `02 Learn_one_tool_at_a_time/Summarize.yxmd` hits.

## Feasibility: why not the exact CTE rewrite

The exact rewrite is two-stage — count per `(group, value)` in a CTE, then take the highest count:

```sql
WITH value_counts AS (
    SELECT grp, val, COUNT(*) AS n FROM src WHERE val IS NOT NULL GROUP BY grp, val
)
SELECT grp, array_agg(val ORDER BY n DESC, val ASC)[1] FROM value_counts GROUP BY grp
```

This **cannot** be built from `sqlalchemy_functions.py`. An `@compiles` hook fills a single aggregate slot in a SELECT list; it never sees the enclosing query's GROUP BY keys, which are assembled independently in `sql_expression.get_select_query` (`plaidcloud/utilities/sql_expression.py`, the `if aggregate:` block that builds `grouping_columns`). Restructuring the statement into a CTE + join-back belongs there, and it would need one extra CTE per mode column so that mode can coexist with the other aggregates in the same SELECT. **Not attempted in this PR** — flagged for a follow-up if exactness beyond the bound below is ever needed.

## What this PR emits instead

Single-expression `approx_top_k` form, **verified executing live on StarRocks 4.1.3**:

```sql
array_filter(
    array_map(sr_mode_e -> sr_mode_e.item, approx_top_k(col, 2, 100000)),
    sr_mode_i -> sr_mode_i IS NOT NULL
)[1]
```

- **Exact**, not approximate, for any group with fewer than 100000 distinct values. StarRocks: *"Expressions that have fewer than `counter_num` distinct items will yield exact item counts"*, and 100000 is the documented maximum `counter_num`. Counters are allocated as distinct values are encountered, so a low-cardinality column pays nothing for the high ceiling. Above 100k distinct values in one group it degrades to a Space-Saving estimate (count error up to `2.0 * numRows / counter_num`), where Databend's `mode()` is always exact.
- **NULLs.** `approx_top_k` counts NULL as its own item and will return it as the winner — verified live — where Databend and Alteryx ignore nulls. `k=2` plus the `IS NOT NULL` filter fixes that: NULL is at most one item, so the top two always contain the most frequent non-null value if the group has one. A group with no non-null value yields NULL.
- **`array_map` is not decoration.** StarRocks' analyzer rejects struct field access applied straight to an aggregate's subscript: `approx_top_k(v, 2, 100000)[1].item must appear in the GROUP BY clause`. The items have to be projected out first.

### Tie-break divergence from Alteryx

Ties go to `approx_top_k`'s internal counter order — arbitrary, but deterministic for a given input. Alteryx's Summarize → Mode returns the **first-encountered** value on a tie; Databend's `mode()` leaves it unspecified. No two of the three targets agree on ties. Documented in the docstring and CHANGELOG.

The `mode() WITHIN GROUP (ORDER BY col)` ordered-set spelling arrives with no clauses — nothing to count — and still raises a clear `CompileError` on StarRocks. It is unchanged on every other dialect.

## Also: median is now exact

`median` did **not** fail closed — it already compiled to `percentile_approx(col, 0.5)`, a t-digest estimate, so a StarRocks workspace could return a different median than a Databend one for identical data. It now compiles to `percentile_cont(col, 0.5)`.

`percentile_cont` linearly interpolates the two middle values on an even count — exactly what both Alteryx's Summarize → Median and Databend's `median()` do. `percentile_disc` was rejected: it returns an actual member of the input (`3`, not `2.5`, for `[1, 2, 3, 10]` — verified live), matching neither. `percentile_cont` also accepts DATE/DATETIME, not just numerics as `percentile_approx` does, so median over a date column works now too. Being exact it holds the group's values rather than a sketch, so a median over a very large group costs more memory than the old estimate did.

## Databend unchanged

Both handlers are `@compiles(..., 'starrocks')` only. Databend/default renders `mode(c)` and `median(c)` byte-for-byte as before; tests assert it, including the ordered-set `mode() WITHIN GROUP (ORDER BY c)` form.

## Live verification (StarRocks 4.1.3)

The exact statement the compiler emits, run end to end:

```sql
SELECT src.g,
       array_filter(array_map(sr_mode_e -> sr_mode_e.item, approx_top_k(src.v, 2, 100000)),
                    sr_mode_i -> sr_mode_i IS NOT NULL)[1] AS mode_val,
       percentile_cont(src.n, 0.5) AS med
  FROM (VALUES ('a','x',1),('a','x',2),('a','y',3),
               ('a',NULL,4),('a',NULL,5),('a',NULL,6),('b',NULL,7)) src(g, v, n)
 GROUP BY src.g
```

→ `a | x | 3.5` and `b | NULL | 7`. `x` wins even though NULL occurs 3 times and `x` only twice, which is the null-skip working; `3.5` is the interpolated median of `1..6`. Also verified wrapped in `CAST(... AS STRING)` / `CAST(... AS DECIMAL(38,10))`, the shape `process_fn` actually produces, and over INT and DATE columns.

## Tests

`plaidcloud/utilities/tests/test_sqlalchemy_functions.py` — StarRocks renderings for `mode`/`median`, the StarRocks arity guard, and Databend byte-unchanged assertions for `mode`, `median` and `mode() WITHIN GROUP`.

Full suite `plaidcloud/utilities/tests/`: the failure set is **identical** to `origin/master` (62 pre-existing failures locally, all from the Snowflake dialect not being installed in the local venv); +3 net new passing tests.


inviscid (commented):
✅ Independent verification — **APPROVE**

mode: `array_filter(array_map(e -> e.item, approx_top_k(col,2,100000)), i -> i IS NOT NULL)[1]` — EXACT below 100k distinct/group (documented StarRocks guarantee), with two live-probed constraints correctly handled: approx_top_k counts NULL as an item (hence k=2 + IS NOT NULL filter) and struct-field access on an aggregate subscript does not parse (hence the load-bearing array_map). Live-verified end-to-end on StarRocks 4.1.3 incl. CAST-wrapped shapes over INT/DATE/DECIMAL.

median: found NOT failing closed but silently compiling to percentile_approx (t-digest ESTIMATE) — meaning StarRocks and Databend could return different medians for identical data. Now percentile_cont (exact, interpolates, matches Alteryx + Databend); percentile_disc correctly rejected (returns a member, not the interpolated midpoint). Memory tradeoff noted.

Databend/default byte-unchanged (both handlers starrocks-only, 3 assertions). Feasibility honestly reported: the exact two-stage CTE is not expressible at the @compiles layer and would belong in get_select_query — not forced. Known divergence: tie-break is arbitrary-but-deterministic vs Alteryx first-encountered (documented).

PR #352, opened 2026-07-28
frame_join_multi: accept dynamic table references for source and target

inviscid:
**This is the one PR in the set that changes behaviour.** The other three (client, plaid, docs) are drafts blocked on a release of this one — see *Rollout* below.

### The problem

`frame_join_multi` required `sources[].source` and `target_frame` to be a bare identifier:

```python
_TABLE_ID_RE = re.compile(r'[A-Za-z0-9_][A-Za-z0-9_-]{0,127}')
```

So a step configured with the table picker's **Dynamic** mode — a name, a project path, or either carrying `{variable}` tokens — could not be saved. `frame_join_multi` is the only step with a save-time validator at all (`_validate_step_on_save` dispatches on that op and no other), which is why no other step has this problem.

### Why relaxing it is safe

The executor has *always* resolved these. `transform_handler.get_frame()`: canonical `analyzetable_` id used as-is; otherwise `apply_variables()`, then `lookup_by_full_path` when the reference contains `/`, else `lookup_by_name` — and `create_if_missing` for the target. `frame_join_multi.py` already calls it for both fields (lines 94 and 240). Only the save-time gate disagreed with the executor.

The gate's stated rationale was keeping dots, quotes and whitespace out of the emitted SQL. That doesn't hold: **the reference never reaches SQL.** `get_frame` returns the canonical id from the RPC's own `table()` response and the query is built from that (`get_table_rep(resolved_id, …)`); the executor says so in its own comment. The lookups on the way are project-scoped RPC calls doing in-memory name comparison, not string-built SQL.

`{variable}` grants no new reach either. `apply_variables` runs strict, and `Formatter().parse()` yields the *whole* field name — so `{x.__class__…}` lands in `bad_keys` and raises before `str.format` is reached. Substitution is a single flat insertion, so an attacker-influenced variable *value* gains nothing. Both `get_frame` calls in this step pass no `project_override`, so lookups stay in the caller's own project.

### What is still rejected

A reference that cannot resolve under any variable values: non-string, empty/whitespace-only, over 512 characters, control characters, malformed braces. That last one otherwise raises inside `apply_variables`' `Formatter().parse()` mid-run, so catching it at save turns a run failure into a message on the step form.

It's a syntax gate, not a resolution gate — a well-formed reference naming a variable that doesn't exist still fails at run time, by design.

Reason tokens are unchanged (`source_not_table_id`, `target_frame_invalid`), so the client's reason-to-tab routing needs no update.

### Tests

Two fixtures repurposed, one deleted. `invalid_source_with_format_chars` (`tab{secret}`) encoded the idea that braces are an injection vector — they're the variable syntax, and the real protection is `apply_variables`' strict check, so it's replaced by a positive `valid_dynamic_table_references` fixture. `invalid_source_not_table_id` now carries an unbalanced brace, which *is* still invalid. Source and target each get accept/reject tables covering id, name, path, variables, and every rejection class.

67 pass in `test_frame_join_multi_validator.py`. (The repo has 25 pre-existing Snowflake failures in `test_sqlalchemy_functions.py` on master, unrelated and untouched.)

### Rollout

`plaidcloud-utilities` is pinned in two data-plane repos and the validator runs in both — save-time in `plaid`, execute-time in `workflow-runner`. **Both are already on `1.22.0` (with `plaidcloud-rpc[full,tracing]==1.13.0`), which is exactly `master` here**, so this PR becomes `1.23.0` and both need the same one-line bump to pick it up.

They must move together: if only the save-time pin advances, a dynamic reference saves cleanly and then fails at run when the runner re-validates with its older copy.

Sequence: merge this → publish a GitHub Release (that's what pushes to PyPI) → bump `plaidcloud-utilities` to `1.23.0` in `plaid` and `workflow-runner` → mark the client, plaid and docs drafts ready.


inviscid (commented):

Reviewed independently, tracing the path rather than trusting this PR's description.

The central security claims hold. The raw reference is resolved to a canonical id via RPC (`table['id']`) before any SQL is built — `get_table_rep` receives only the resolved id. `apply_variables`' strict `bad_keys` check compares the full dotted `field_name` from `Formatter().parse`, so `{x.__class__}` is rejected as an undefined key before `str.format` is reached. Neither `get_frame` call in `frame_join_multi.py` passes `project_override`, so lookups stay in the caller's own project.

The decisive point: `frame_extract` already calls the same `get_frame()` with **zero** save-time validation. Path and variable resolution is not new surface — this brings one step's gate in line with what the platform already does unguarded everywhere else.

Verified by direct testing rather than inspection: `_is_valid_table_reference`'s brace check only ever raises `ValueError` (fuzzed a dozen brace patterns, consistent with the single `except`); nothing else references the removed `_TABLE_ID_RE`; reason tokens are unchanged, so no client update is needed; and the local `test_plaidcloud_dtype_aliases_accepted` failures reproduce identically on unmodified `master`, so they're pre-existing and unrelated.

One nuance confirmed and judged non-blocking: a nested replacement field in a format spec (`{a:{b.attr}}`) escapes the top-level `bad_keys` check, and `str.format` does resolve it. It reaches only attributes of variables the same principal who authored the step already controls, it is a latent property of `apply_variables` shared by every step rather than anything introduced here, and the validator's own comment already discloses the scenario.

Test coverage for the new boundary is thorough — id, name, path, variables, crossed with every rejection class, at both `source` and `target_frame`.

No findings.

PR #353, opened 2026-07-29
Preserve binary external source columns

inviscid:
## Summary
- Leave largebinary source projections uncast so databases return their native binary values.
- Cover SQL Server dialect compilation to ensure no CAST is generated.
- Document SQL Server rowversion support.

## Validation
- python3 -m py_compile plaidcloud/utilities/sql_expression.py plaidcloud/utilities/tests/test_sql_expression.py
- git diff --check
- pytest plaidcloud/utilities/tests/test_sql_expression.py -q (blocked locally: installed plaidcloud-rpc lacks PlaidCurrency and PlaidUnicode; CI has the locked dependency set).

## Follow-up
Release this package, then update Plaid's plaidcloud-utilities pin before deploying the coordinated staging fix.


inviscid:
APPROVED

PR #354, opened 2026-07-29
sql_expression: add get_update_rewrite_query (projection form of get_update_query)

inviscid:
Root of a five-repo set reactivating **Table: Update** as a step that writes to a target table. Nothing else can land until this releases.

## What

`get_update_rewrite_query` — the projection form of `get_update_query`, for update steps that write to a target table instead of mutating their source. StarRocks rejects `UPDATE` on Iceberg tables outright (`table … does not support update`), which is why the step was pulled from the New Step menu; the same outcome is reachable as a plain `SELECT` provided the filter is *translated* rather than reused.

```sql
-- get_update_query
UPDATE t SET amount = amount * 2 WHERE region = 'WEST'

-- get_update_rewrite_query
SELECT id,
       region,
       CASE WHEN region = 'WEST' THEN amount * 2 ELSE amount END AS amount,
       note
FROM   t
```

The wheres pick which rows get **updated**, not which rows come back, so they become a `CASE` condition per mapped column and every source row is projected — mapped columns conditionally, unmapped columns as passthroughs.

## Why `CASE` and not two filtered branches unioned

`WHERE f` / `WHERE NOT f` is not a partition. A row where `f` evaluates to NULL satisfies neither and vanishes. Verified live on StarRocks with a 6-row fixture containing two NULLs: **the union returned 4 rows.** `CASE` falls through to `ELSE` for FALSE *and* NULL, which is exactly `UPDATE … WHERE` semantics, and it scans once.

## Notes for review

- **Value rules are shared, not reimplemented.** Both functions drive off the same `get_update_value`, so `nullify` / `constant` / `expression` and the text-dtype-blank quirk cannot drift between the two query shapes. A non-`ClauseElement` value is wrapped in a typed literal so an unconditional `nullify` still carries its column's type.
- **One deliberate addition.** `update().values()` rejects a key that isn't a column of the table; a projection has no such check, and losing it would turn a stale mapping into a step that reports success having updated nothing. An update column absent from the source now raises `SQLExpressionError` naming it. A config with **nothing** to update — every update column missing a constant, an expression and Set Null — is rejected the same way: the equivalent `UPDATE … SET` with no assignments was invalid SQL and always failed loudly, whereas its projection is a valid query (an exact copy of the source) that would otherwise report success having changed nothing.
- `get_update_query` is untouched and remains the path for Databend, Postgres and Greenplum.

## Tests

10 new, 163/163 green. Covers: filter→`CASE`, an explicit assertion that **no `WHERE` is emitted**, the column-count-and-order invariant, unconditional value with no filter, `nullify` with and without a filter, the text-`''` quirk under a filter, non-text passthrough, the unknown-column raise, `{variable}` substitution, and spaced/dotted column names (`"Calendar Date"`, `"1.0"` — the `frame_pivot` dot-column regression class).

## Companion PRs

Consumed by workflow-runner. Both repos need their pin bumped to the release cut from this branch.

- PlaidCloud/workflow-runner — runner transform
- PlaidCloud/plaid — step metadata + migration report
- PlaidCloud/PlaidClient — step form + menu
- PlaidCloud/plaidcloud-docs — reference page + release notes


inviscid:
**Adversarial review round 1 — one finding, fixed (force-pushed into the commit).**

**A config with nothing to update compiled to a silent full-table copy.** The old `get_update_query` produced `UPDATE t SET` (no assignments) for that shape — invalid SQL, so it always failed loudly. The projection form of the same config is a perfectly valid query: an exact copy of the source. The step would have created the target, inserted an unmodified clone, and reported success.

The reviewer proposed requiring `target_columns` in the runner. That's incomplete — it misses the shape that actually matters, where `target_columns` is *non-empty* but every entry is excluded (e.g. a single numeric column with no constant, expression or Set Null, which `get_update_value` drops). The guard is instead at the value level, here, so it covers both:

```python
if not new_values:
    raise SQLExpressionError('No columns to update: …')
```

Also switched the unknown-column raise from `ValueError` to `SQLExpressionError`. The runner's `handle_error` special-cases `SQLExpressionError` and shows the message; a bare `ValueError` falls to the generic branch and surfaces a traceback instead. `SQLExpressionError` is already what this module raises for user config errors.

Tests: 164 green. `test_nothing_to_update_raises` covers both shapes.

**Reviewed and deliberately not changed:** the target-vs-source guard ordering in the runner (verified non-exploitable and identical to `frame_extract`'s established pattern — a config-level pre-check would catch only a subset, since it can't see name-vs-id aliasing).

No issues found in the core projection logic: filter→`CASE` never `WHERE`, NULL-condition parity with `UPDATE … WHERE`, column order/completeness, alias-vs-Table handling, and literal type coercion all verified.

inviscid (commented):

Verified by generating and comparing compiled SQL from both `get_update_query` and `get_update_rewrite_query` across filtered/unfiltered, `nullify`, `constant`, `expression`, the text-dtype-blank quirk, unmapped-column passthrough, unknown-column, and empty-update-set cases — the `CASE WHEN cond THEN val ELSE col END` output matches `UPDATE … SET col=val WHERE cond` semantics in every case, including the NULL-condition fall-through to `ELSE`. Value logic is shared via the untouched `get_update_value`, so the two query shapes cannot drift, and `get_update_query` itself is unmodified. The new `SQLExpressionError` guards are sensible, documented, and match the codebase's existing exception convention (workflow-runner catches `SQLExpressionError` and surfaces it as a user error). Tests use real compiled-SQL equivalence rather than a structural comparison, so they are not gameable.

No correctness, security, or contract issues found. The one nit raised — the PR description saying `ValueError` where the code says `SQLExpressionError` — has been corrected in the description.

PR #355, opened 2026-07-31
fix(sql_expression): tolerate a null target-column dtype (sc-23460)

simozzy:
Layer 1 of 3 for [sc-23460](https://app.shortcut.com/plaidcloud/story/23460) (Sev-1, Hyster-Yale). This is the chokepoint fix, and **the only one that rescues steps that are already saved broken**.

## The bug

`sqlalchemy_from_dtype` stringifies its argument before the regex lookup, so a target column carrying `dtype: null` becomes the key `'none'` and raises `RegexMapKeyError` instead of returning a type:

```
File "plaidcloud/utilities/sql_expression.py", line 710, in get_from_clause
  cast_type = sqlalchemy_from_dtype(target_column_config.get('dtype'))
File "plaidcloud/rpc/functions.py", line 165, in lookup
  raise RegexMapKeyError(key)
plaidcloud.rpc.functions.RegexMapKeyError: 'none'
```

That killed two things: the **save** (a 500 behind the client's "Update Failed – Internal Server Error", via plaid's REST expression validation) and the **step run** (`workflow_runner`'s `frame_lookup` calls the same `get_select_query`).

Null-dtype columns reach here because PlaidClient's YAML→JS step-form conversion dropped the `default: text` the column definitions carried, so a hand-added output row was written with no type. That is fixed separately in PlaidClient — but only for rows added *after* it ships. Rows already persisted without a type stay broken until this lands.

## The fix

`_target_dtype()` resolves the dtype rather than passing it through raw: **configured value → matching source column's dtype → `text`**.

Falling back to the source column rather than straight to `text` is the part that matters — a null-dtype `amount` column now compiles to `CAST(... AS DECIMAL(38, 10))`, not `VARCHAR`, so a summed numeric column isn't silently stringified. `text` is the last resort, and is exactly what the step forms' own column default has always been.

Two details:

- The `largebinary` no-cast check in `source_from_clause` resolves through the same helper, so a null-dtype binary column still isn't cast.
- The source lookup matches the configured name first, then the name with a table qualifier stripped, so a column genuinely named `a.b` wins over the `b` of a `table.b` reading. It is deliberately **not** alias-aware — a qualified source takes the first list carrying that bare name, which can be the wrong side when both tables share a column name. That only applies where the config previously raised outright, and lands no worse than the forms' own `text` default, so threading alias resolution through here isn't worth it.

A dtype that is set but genuinely unrecognised still raises, as before.

## Tests

Four new cases in `TestGetFromClause`: source-dtype fallback, table-qualified source, `None`/`''` without a source falling back to text, and a null-dtype largebinary source still not casting. Full suite: **727 passed, 2 skipped**.

## Downstream

`plaid` needs a release of this plus a `requirements.txt` pin bump (currently `plaidcloud-utilities==1.25.0`) to get the already-saved-step rescue. The plaid-side fix (layer 2) unblocks *new* saves without it, so the customer isn't waiting on this release — but the run-time failures are.

Related: PlaidCloud/plaid (layer 2), PlaidCloud/PlaidClient (layer 3).


simozzy:
## Adversarial review round 1 — no blockers, one nit found and fixed

Recording the review outcome here for parity with the sibling PRs (PlaidCloud/plaid#6847, PlaidCloud/PlaidClient#1939).

### Verified, not assumed

The reviewer re-ran the suite and read `origin/master` rather than reasoning from the diff:

- **No existing step's SQL can change.** `_target_dtype`'s `if dtype: return dtype` fires on any truthy configured value and returns it verbatim, so a config whose dtype is already set compiles exactly as before. This was the main regression risk and it is closed by construction, not by test coverage alone.
- **`largebinary` is unaffected** for already-working configs — an existing largebinary dtype takes the same early return, so `source_from_clause`'s no-cast check behaves identically to the old raw read.
- **Nothing relied on the old raise.** No test in the repo asserts `RegexMapKeyError`, and the prior CHANGELOG line stating that a missing dtype "still raises upstream" is explicitly superseded by this entry.
- **All four original tests genuinely fail on master** — master computes `cast_type` unconditionally at the top of `get_from_clause`, before any branch, so `dtype: None` always raised regardless of constant/expression/source shape. None of the tests are vacuous.

### Both seed findings came back negative, with reasoning I accept

- **Raw `dtype` reads left in the `serial`/`bigserial` and `MAGIC_COLUMN_MAPPING` branches** are not an inconsistency to clean up. `None` can never satisfy either membership test, and those lines are only reached when `source`, `constant` and `expression` are all falsy — a shape that now falls through to the unchanged, pre-existing `SQLExpressionError` about needing a Constant, an Expression or a Source Column. The same raw-read-for-sentinel-detection pattern appears elsewhere in the file, so leaving it alone is the consistent choice.
- **The two `_target_dtype` calls cannot disagree.** `get_from_clause` passes the identical config and source-column objects into `source_from_clause`; neither mutates them, and the function is pure.

### The nit was better than my docstring, so I changed the code (`54cc5a6`)

The fallback returned the **first** matching list's dtype for a name carried by more than one source table, even when the target explicitly qualified the other side — `table2.Amount` resolving to table1's `numeric` when table2's is `text`. My docstring waved this away as landing "no worse than the forms' own text default". That was wrong, and the distinction matters: `CAST(<text column> AS NUMERIC)` can be rejected by the warehouse, whereas casting a numeric column to text cannot fail. Those are not equivalent outcomes.

A name whose candidate dtypes disagree is now **ambiguous and falls through to `text`** rather than picking a side; a name every side agrees on still resolves to it. Two tests pin the tie-break so a later refactor can't change it silently. **729 passed, 2 skipped.**

### Cross-layer note — this guard was dead until plaid agreed with it

Worth recording because it is not visible from this diff alone: `plaid`'s `get_simple_expression_sql` back-fills a missing dtype **before** the column reaches `_target_dtype`, so on the save path plaid wins and the resolution here never runs. The ambiguity guard above was therefore unreachable for every save until the same rule was mirrored plaid-side (PlaidCloud/plaid#6847, `93064c401d`). Any future change to the resolution rule in either repo has to be made in both, or one of them becomes dead code.

**Open, being checked now:** a round-2 review on the plaid side is examining a suspected divergence between the two — plaid does not filter falsy dtypes when building its candidate set, whereas `_target_dtype` here does (`and sc.get('dtype')`). If confirmed, `{None, 'numeric'}` would read as ambiguous plaid-side and resolve to `numeric` here. I'll fix whichever side is wrong once that lands; flagging it rather than leaving it implicit.

### Scope this PR does *not* cover

Releasing this does **not** by itself let an already-saved null-dtype step run again. There are two independent crash points, and this fixes one:

1. The SQL build — fixed here.
2. The **table-shape write** — `workflow_runner`'s `frame_lookup` passes `target_columns` straight into `analyze.table.touch(meta=…)`, and plaid's `datastore_tools.create_table` → `col_type()` hands the `None` on to `sqlalchemy_from_dtype`. Not fixed; deliberately deferred (that helper is on the path of every table creation) and documented in plaid's `CLAUDE.md`.

Also relevant to release planning: **`plaid` and `workflow-runner` pin `plaidcloud-utilities` independently** (`requirements.txt` and `requirements-base.txt`, both `1.25.0` today). Bumping only plaid's leaves every step *run* on the old library.

simozzy:
## Merged with master, plus one non-regressive fix to the fallback itself

`mergeable` was `CONFLICTING`; the only conflict was `CHANGELOG.md` (both sides added a bullet at the top of `## Unreleased`) — resolved as a union, sc-23460 alongside sc-23414. The code auto-merged: master's change since the branch point was #356's `get_update_query` empty-update rejection, a different function. **734 passed, 2 skipped** on the merged tree; `ruff` reports nothing new on either changed file (the 8 findings on them are pre-existing, and the job is `--exit-zero` anyway).

---

### The finding: the fallback declined to resolve cases the module already resolves

`_target_dtype` scanned every source's columns as one flat list, so a column name carried by more than one source looked ambiguous and fell through to `text`. But `source_from_clause` doesn't treat those as ambiguous — it calls `get_column_table`, which resolves the reference to exactly one table by a documented precedence (`source_alias` → legacy `source_table` → positional `tableN.col` → name-intersect), and emits *that* table's column.

So the two disagreed, and the fallback lost in the cases it was written for. Measured on a two-source fixture whose `Amount` is text on one side and numeric on the other:

| target config | before | after |
|---|---|---|
| `source: 'table2.Amount'` | `text` | `numeric` |
| `source: 'Amount'`, `source_alias: 'b'` | `text` | `numeric` |
| `source: 'Amount'` (bare, shared name) | `text` | follows the table actually selected |

The first two name their side outright — there is nothing to guess — and both were being stringified. That is the exact harm the PR's own rationale gives for falling back to the source column rather than straight to `text`: *"what keeps a summed numeric column numeric instead of silently stringifying it."* `source_alias` is the `frame_join_multi` path, so this was the multi-table join case in particular.

**The fix narrows the candidate list through `get_column_table` before matching the name**, reusing that precedence rather than re-deriving a weaker copy — so the dtype can't disagree with the column being cast.

Three things keep it non-regressive:

- **It cannot raise where the old helper didn't.** `get_column_table` raises on what it considers unresolvable (an unknown alias, `source_table` with >2 sources, a shared name across >2 sources). Those are caught and fall back to the original flat scan, so the error still surfaces from `source_from_clause` a moment later, with its message, as it does today.
- **Single-source and single-candidate paths are untouched** — the helper short-circuits before doing any resolution when there are fewer than two tables.
- **Genuine ambiguity still lands on `text`.** A shared name with no way to pick a side keeps the existing behaviour, for the reason already recorded on the branch: a wrong guess yields `CAST(<text column> AS NUMERIC)`, which the warehouse can reject, where casting a numeric column to text cannot.

### The ambiguity test wasn't testing what it said

`test_null_dtype_on_an_ambiguous_source_name_falls_back_to_text` asserted `text` for `source: 'table2.Amount'` against a fixture whose *table2* column was itself `text` — so it passed under either behaviour, and would not have caught this. The cases now use fixtures that tell the two apart, and cover the positional reference, the `source_alias` path, the bare shared name, and the genuinely-unresolvable case separately.

### Checked and left alone

- **Other `sqlalchemy_from_dtype` call sites.** `get_update_value` defaults through `dtype_map.get(source, 'text')` and `get_table_rep`/`analyze_table` read `sc['dtype']` off source columns, which the config layer populates — none of them take the target-column config this bug comes from. The chokepoint claim in the description holds.
- **The `largebinary` no-cast check** resolves through the same helper, so it picks up the narrowing too — a null-dtype binary column still isn't cast, now including when the binary source is one of several tables.
- **`CHANGELOG.md`** — the bullet described the old ambiguity rule; corrected to describe the narrowing, since it's the user-visible semantics.

The merge is a separate commit from the fix so the resolution can be read on its own.

rad-pat (approved):
## Approving — the central claim holds under test, with one cross-repo follow-up to book

Reviewed by building the branch out and exercising it, not by reading the diff. Set-up: fresh venv, `pip install -e .[test]`, `plaidcloud-rpc` 1.14.0.

### The invariant the merge commit turns on is real

Commit `4e54260`'s claim is *"the dtype can't disagree with the column being cast."* The four new tests assert `_target_dtype`'s **return value** in isolation, so none of them actually check that — they'd pass even if `source_from_clause` emitted a different table's column. Compiled the real thing against a two-source fixture (`table_1.Amount` text, `table_2.Amount` numeric) and compared the resolved dtype with the table the emitted column came from, across all four `get_column_table` rules:

| target config | resolved | emitted from | |
|---|---|---|---|
| `source: 'table2.Amount'` | `numeric` | `table_2` | ✅ |
| `source: 'table1.Amount'` | `text` | `table_1` | ✅ |
| `source: 'Amount'` (bare, shared) | `text` | `table_1` | ✅ |
| `source_table: 'table2'` (legacy) | `numeric` | `table_2` | ✅ |
| `source_alias: 'a'` / `'b'` | `text` / `numeric` | `table_1` / `table_2` | ✅ |

The cast matches the column in every case. That's the property worth having, and it holds — but it's held by construction, not by the test suite. **Worth one test that compiles the SELECT** rather than five that poke a private function.

### Non-vacuity, checked properly

The four narrowing tests fail on `54cc5a6` only with `TypeError: takes 2 positional arguments but 3 were given` — a signature change, which proves nothing about behaviour. Re-ran them with `_owning_source_columns` monkeypatched back to the old flat scan (new signature, old body). Two fail on value (`'text' != 'numeric'`): the positional and `source_alias` cases genuinely pin the narrowing.

`test_null_dtype_follows_the_table_the_query_reads` **passes under both** — its fixture's first table is `text`, and the flat scan calls a shared name ambiguous and returns `text` too, so the two behaviours coincide. Same shape of problem as the one you caught in round 1. Flipping the fixture (first table numeric) would make it distinguish.

### The regression risk is closed

`dtype` set → `if dtype: return dtype` before anything else runs, so `_owning_source_columns`/`get_column_table` are never reached on the hot path — no added `get_column_table` calls for any existing config. Compiled `numeric`/`text`/`largebinary`/`integer` against master: byte-identical, `largebinary` still uncast. An unrecognised dtype still raises `RegexMapKeyError: 'wibble'`.

The "must never turn a fallback into a raise" contract holds where I could break it:

- 3 sources sharing a name → `except` fires, flat scan, then `source_from_clause` raises `Column 'Amount' exists in multiple sources; set source_alias explicitly` ✅
- unknown `source_alias` → same, raises `source_alias='zzz' not found in sources` ✅
- source name in no config → `text`, then `Cannot find source column Nope in table t` ✅
- `source_column_configs=None`, or shorter than `tables` → `text`, no raise ✅

Also an improvement not claimed: a target column with no source/constant/expression and a null dtype now reaches the real `needs either a Constant, an Expression or a Source Column!` instead of `RegexMapKeyError: 'none'`.

### Merge + gates

`742 passed, 2 skipped` on `origin/master` merged in. Only `CHANGELOG.md` conflicts (both sides at the top of `## Unreleased`); `sql_expression.py` and its tests auto-merge against #357 and #359. Patch coverage under master's new `--fail-under=80` gate (#358) is **90%** — the 3 uncovered lines are `764-765` and `769`, i.e. both `_owning_source_columns` fallbacks. Those are exactly the paths whose "never raise" contract carries the safety argument; I hit them by hand above, but nothing in the suite does.

---

## The one thing to book before this is forgotten

`plaid` master's `resolve_dtype` (merged in #6847, `93064c40`) opens with:

> *Deliberately algorithm-identical to `sql_expression._target_dtype` … They must agree because they are BOTH live, on different paths. If the two rules differ, the type the user was shown when saving is not the type that runs.*

`4e54260` (21:51 on 2 Aug) added the `get_column_table` narrowing **after** #6847 merged (17:29 the same day), and plaid was not updated. That docstring is now false. Ran plaid master's `resolve_dtype` verbatim beside this branch's `_target_dtype` on the same fixture:

| target config | plaid (save) | utilities (run) | |
|---|---|---|---|
| `source: 'table2.Amount'` | `text` | `numeric` | ✗ |
| `source: 'Amount'`, `source_alias: 'b'` | `text` | `numeric` | ✗ |
| `source: 'Amount'`, `source_table: 'table2'` | `text` | `numeric` | ✗ |
| `source: 'table1.Amount'` | `text` | `text` | ✓ |
| `source: 'Amount'` (bare) | `text` | `text` | ✓ |

Not a blocker for this PR, and the asymmetry runs the safe way: **this side is the correct one in all three**, plaid is merely more conservative, and its back-fill only touches a validation deepcopy that is never persisted — so no cast the warehouse can reject comes out of it. But it is precisely the drift your own round-1 note warned about (*"has to be made in both, or one of them becomes dead code"*), and a stale comment asserting a false invariant is worse than no comment. Either port the narrowing to plaid or correct the docstring to say plaid is deliberately the conservative subset.

## Smaller notes, none blocking

- **`_owning_source_columns` matches tables by identity** (`candidate is table`). Fine when `tables_by_alias`'s values are the same objects as in `tables`; when they're equivalent-but-distinct, resolution silently degrades to the flat scan and back to `text`. Confirmed both ways with the executor's own `select(get_table_rep(..., alias=…)).subquery(alias)` shape. Safe degradation — the flat scan only returns a dtype when every source agrees or one carries the name — but the `source_alias` win is contingent on an invariant owned by workflow-runner, and the test asserts it with hand-built identical objects, so it can't catch a violation. A line in the docstring naming that requirement would pay for itself.
- **`except Exception`** is broad, but defensible here for the reason given — a `TypeError` from misaligned configs should be absorbed and re-raised downstream by `source_from_clause`, not converted into a new failure mode by the dtype helper.
- **Adjacent, out of scope:** `get_select_query` line 1231 filters on `tc['dtype']` by subscript, so a target column with the `dtype` key *absent* (not null) still raises `KeyError: 'dtype'` when `use_row_number_for_serial=False`. The null case is fine on every path. Pre-existing, and plaid back-fills absent keys, but the CHANGELOG says "missing or **null**" and "missing" is only true on one branch.

Both siblings are merged (plaid #6847 2 Aug, PlaidClient #1939 today), so this is the last layer. Release still needs **both** pins moved — `plaid/requirements.txt` and `workflow-runner/requirements-base.txt` — or step *runs* stay on the old library.

Approving. Resolve the `CHANGELOG.md` conflict as a union and merge.

PR #356, opened 2026-08-02
sc-23414: reject an empty update in get_update_query instead of emitting invalid SQL

inviscid:
## Symptom

A Table Update step with a **numeric** column dragged into Update Columns and Constant, Expression and Set Null all left blank saves cleanly and fails at run with a database syntax error that names neither the step nor the column.

## Root cause

`plaidcloud/utilities/sql_expression.py:1280` `get_update_query` builds its `SET` list from `get_update_value`, which returns `(False, None)` for a non-text column carrying no `constant`, no `expression` and no `nullify` (`:1276`). When that excludes every mapped column, `values` is `{}` and the function returns `update(table).values({})`, which compiles to:

```
UPDATE anlz_schema.analyzetable_x SET  WHERE anlz_schema.analyzetable_x.id = 1
```

`SET` with nothing after it. Reproduced on this branch's parent commit (`b2ba91d`) with the story's snippet — output byte-identical to the one in the report.

## What changed

`get_update_query` raises `SQLExpressionError` when the values dict comes out empty, naming the update columns:

```
No columns to update: every update column is missing a constant, an expression and Set Null. Columns: ['amount']
```

This mirrors the guard the sibling `get_update_rewrite_query` already carries for the same config (`:1323-1330`). A silent no-op would be worse — the config is genuinely meaningless.

Plus a CHANGELOG entry and two tests.

## Scope / blast radius

- **Text columns unaffected** — `get_update_value` returns `''` for a text column with no value, so it stays in the `SET` list. Covered by the existing `test_empty_string_no_matter_what`.
- **Table Delete unaffected** — `get_delete_query` (`:1349`) maps no columns and cannot produce an empty `SET`. Verified: it still compiles to `DELETE FROM … WHERE …`. The story title's "/Delete" is inaccurate.
- **No runtime callers.** `git grep -nE 'get_update_query' <origin default branch>` across every repo in `~/Projects` returns only the definition, the CHANGELOG and this repo's tests; workflow-runner `origin/master` and `origin/beta-202607` both call `get_update_rewrite_query` (moved in #1148). The behaviour change therefore reaches only tenants pinned to a pre-#1148 executor, for which `get_update_query` remains the Databend / Postgres / Greenplum path — and there the old behaviour was a hard DB error anyway, so nothing that worked stops working.

## Verification

```
uv venv .venv --python 3.12 && uv pip install -e '.[test]'
pytest --ignore=plaidcloud/utilities/xlwings_utility.py
```
→ **758 passed, 2 skipped, 463 subtests passed** (the `--ignore` matches what `conftest.py` does on Linux/CI; `xlwings` is Windows-only in `requirements.txt`).

Fails-before confirmed by writing the tests first and running them against unmodified source: `AssertionError: SQLExpressionError not raised` on both new tests, 4 pre-existing `TestGetUpdateQuery` cases still passing.

`ruff check` on both touched files produces no findings that weren't already there on master.

Story: https://app.shortcut.com/plaidcloud/story/23414


inviscid (commented):

Independent review confirms the PR's claims hold up. The guard mirrors `get_update_rewrite_query`'s, and the reviewer verified independently that `get_update_query` has zero live callers — GitHub code search across the PlaidCloud org finds no references outside plaid-utilities itself, and `workflow-runner/.../frame_update.py` (the sole Table Update executor) calls only `get_update_rewrite_query`, unconditionally for all engines.

The new tests genuinely exercise the added path: `setUp` maps `Column2`/`Column3` as `numeric`, so both tests hit `get_update_value` returning `(False, None)` for every target column, and would fail against unmodified source with `AssertionError: SQLExpressionError not raised`.

Non-blocking findings, recorded and consciously accepted:

- `plaidcloud/utilities/sql_expression.py:1327` vs `:1301` — the new guard's message appends `Columns: [...]` where the sibling `get_update_rewrite_query` guard does not. Cosmetic inconsistency; the columns-list version is arguably the better of the two.
- PR body ("Scope / blast radius") — overstates that `get_update_query` "remains the Databend / Postgres / Greenplum path" for pre-#1148 pinned tenants, implying a live per-engine fallback. Current `frame_update.py` has no such branch at all; it is fully switched to the rewrite form. Description precision only — the underlying "zero callers today" claim is correct.

CI: 11 checks, 0 failures. `Perform Tests` (pytest) ran and passed inside the lint workflow; 758 passed / 2 skipped / 463 subtests locally.

PR #357, opened 2026-08-02
fix(sql_expression): reject a rule condition that compiles to no SQL (sc-23440)

inviscid:
Companion guard for [sc-23440](https://app.shortcut.com/plaidcloud/story/23440). The generator that was producing the bad condition is fixed in [workflow-runner#1156](https://github.com/PlaidCloud/workflow-runner/pull/1156); this makes the failure impossible to reach silently from any caller.

## What was wrong

`apply_rules` filters rules on the condition *text* being non-null and non-empty:

```python
valid_rules = df_rules[... & (df_rules[condition_column].notnull()) & (df_rules[condition_column] != '')]
```

but the predicate actually placed in the query is `eval_rule(...)` of that text, and nothing checked what *that* produced. A condition of `and_()` is non-empty text that compiles to a zero-clause `BooleanClauseList`, and the two paths then fail in opposite ways:

| `include_once` | result |
|---|---|
| on | `case()` emits a branch with no condition (`WHEN  THEN 'C0005'`); the warehouse rejects the statement |
| off | `select().where(<empty>)` — SQLAlchemy **drops the WHERE entirely**, so the rule matches every source row and the run completes, quietly wrong |

Guarding only the `case()` branch would have left the worse half in place, so both go through the same check.

## The change

`rule_predicate()` compiles each condition once and raises `SQLExpressionError` naming the offending `rule_id` and its condition text if the result renders to nothing.

Two details behind the check:

- **`str(predicate)` on the unbound expression** is the general emptiness test — `''` for a zero-clause `and_()`, non-empty for any real predicate — and needs neither a dialect nor a bind context. A `None` predicate (condition text `null`) is rejected explicitly, since `str(None)` is the non-empty `'None'` and would otherwise reach a `WHEN NULL THEN` that never matches.
- **Any failure to render means "not empty", never "empty".** `str()` compiles under the *default* dialect, and a construct whose compiler is dialect-specific raises there — with whatever it likes. `titlecase` raises `CompileError`; `slice_string` raises `NotImplementedError`. With 71 dialect-specific compilers in `sqlalchemy_functions`, guessing at exception types is the wrong shape, so the probe catches broadly. This is safe because the probe asks exactly one question and a failed render is never an empty one; a genuinely broken expression still fails at execution, against the dialect that has to run it and with a message about that dialect. Note the `try` wraps *only* the render probe — `eval_rule` is called outside it, so a real evaluation error still propagates.

A genuine catch-all (`and_(true)`) is accepted, which is what the workflow-runner PR now generates.

## Why keep it once the generator is fixed

It is defense in depth for that case, it stands on its own for any other condition that compiles away, and it converts the current 40 000-character parse error into a one-line diagnosis for a tenant running an older workflow-runner against this release.

## Tests

`TestApplyRules` — 6 tests. Three fail on `master` and pass here: the empty predicate under `include_once` both ways, and the `null` condition. The other three pass on `master` by construction and pin no-regression — the catch-all is accepted, and neither the `CompileError` nor the `NotImplementedError` construct is mistaken for empty. Those last two cannot fail on `master`, which has no render probe at all; the `NotImplementedError` case was verified by mutating the shipped `except Exception` back to `except sqlalchemy.exc.SQLAlchemyError`, under which it fails.

Suite: 475 passed excluding `test_sqlalchemy_functions.py` and `test_sqlalchemy_views.py`, whose 25 Snowflake-dialect failures reproduce unchanged on `master` and are untouched by this diff. Note this repo's CI runs **lint only — no pytest** — so that figure is from a local run. Ruff: the 8 findings in the touched files are byte-identical on `master`; this diff adds none.

## Release

Needs to go out as **1.26.0** before workflow-runner can bump its `requirements-base.txt` pin off 1.25.0. Note `master` already carried an `Unreleased` block, so that release ships more than sc-23440.


inviscid (commented):

Independent Sonnet reviewer, fresh context, given the repo's reviewer guidelines. It was pointed specifically at the broad `except Exception` and asked whether catching broadly is justified or masks real errors.

Its findings: the guard closes a real gap — `valid_rules` filters on condition *text*, not on what the compiled predicate renders to. It verified against SQLAlchemy 2.0.50 that `str(sqlalchemy.and_())` is `''`, confirming the premise. It confirmed the `try` is scoped correctly (`eval_rule` runs outside it; only the `str(predicate)` render probe is guarded) and judged the broad catch deliberate and justified: with 71 dialect-specific compilers in `sqlalchemy_functions` raising arbitrary types under the default dialect, any failure to render is provably "not empty", and a genuinely broken expression still fails at real execution against the dialect that has to run it. It ran `TestApplyRules` against both master and this branch: 3 of 6 fail on master and all 6 pass here, confirming the tests are not vacuous.

No blocking issues.

Non-blocking finding, now fixed:

- The PR description claimed 4 tests fail on master; only 3 do. The `NotImplementedError` render test cannot fail on master, which has no render probe at all — it fails against the intermediate narrow-`except` commit, which is how that regression was actually verified. PR body corrected before merge.

PR #358, opened 2026-08-02
ci: gate PRs on 80% patch coverage; fix the non-gating global coverage gate

inviscid:
> ### ⚠️ This gate is NOT enforced by branch protection today
>
> **`plaid-utilities` `master` has zero required status checks.** Verified:
>
> ```
> $ gh api repos/PlaidCloud/plaid-utilities/branches/master/protection
> required_status_checks.contexts = []      ← EMPTY
> required_status_checks.checks   = []      ← EMPTY
> required_status_checks.strict   = true    (nothing to be strict about)
> enforce_admins                  = false
> required_approving_review_count = 1
>
> $ gh api repos/PlaidCloud/plaid-utilities/rulesets   →  0
> ```
>
> This is not specific to the new gate. The existing `Lint PlaidCloud Utilities w ReviewDog` job — the one that runs `pytest` — **is not required either**. A PR can merge on this repo today with tests outright failing, given one approval.
>
> So this PR **establishes a job that fails**, which is weaker than an enforced gate. To make it actually bite, `Lint PlaidCloud Utilities w ReviewDog` must be added to the required status checks for `master`.
>
> **Owed follow-up, deliberately not done here:** changing branch protection alters merge rules for everyone on the repo and is the owner's call. No protection settings were touched by this PR.

## Why

Two problems, one of which is that the existing coverage gate does nothing.

### 1. The current gate is inert

`.github/workflows/pr_lint.yaml` sets `minimum_coverage: 75` on `5monkeys/cobertura-action`, but that action's `fail_below_threshold` input **defaults to `false`**:

```yaml
  fail_below_threshold:
    description: 'Fail the action when the minimum coverage was not met.'
    required: false
    default: false
```

With it off, the action reports the GitHub `neutral` conclusion, and **neutral does not fail a PR**. Verified against merged PR #356:

```
$ gh pr checks 356
coverage    skipping    0    ...

$ gh api .../check-runs
coverage | status=completed | conclusion=neutral
Lint PlaidCloud Utilities w ReviewDog | status=completed | conclusion=success
```

That check run's own summary reads `**All files** | 57% | :x:` with the footer `Minimum allowed coverage is 75%`. So the repo has been 18 points under a threshold that annotates files with crosses and gates nothing, and #356 merged green.

(Note this was *doubly* inert: neutral wouldn't have failed the PR, and even a hard failure wouldn't have blocked the merge, per the branch-protection banner above.)

### 2. Global coverage is the wrong instrument here

57% overall is historical debt, not any given PR's quality, and a large refactor is in flight. Gate the lines the PR actually changed.

## What changed

| Change | Why |
| --- | --- |
| `actions/checkout` → `fetch-depth: 0` | Depth-1 clone has **no merge-base** with the base branch. Without this diff-cover reports "No lines with coverage information in this diff" and exits **0** — a false pass, not a pass. |
| New `Check Patch Coverage` step, `diff-cover==10.4.1`, `--fail-under=80` | Gates coverage of changed lines only. Pinned, not floating. |
| Compare branch `origin/${BASE_REF}` | `github.base_ref` passed via `env:` rather than interpolated into the shell body. |
| `--format markdown:"${GITHUB_STEP_SUMMARY}"` | Result is visible on the run page instead of buried in the log. `--format` replaces the now-deprecated `--markdown-report`; console output is still printed. |
| cobertura step: `fail_below_threshold: false`, `minimum_coverage: 75` → `57` | See below. |
| `get_update_query`: nested comprehension → plain loop | The deliberate covered source change, so the gate has real lines to measure. |

### `if: success()` vs `if: !cancelled()`

The gate carries **neither** — it takes the default, which *is* `success()`, so writing it out would be a no-op. That is deliberate and commented:

- The neighbouring `Test Report` / `Prepare Coverage Report` steps use `success() || failure()` because you want the annotations even on a red run.
- The **gate** must not. A failed `pytest` leaves `coverage.xml` stale or absent, so diff-cover would compute a meaningless number and stack a second, misleading failure on top of the real one. Skipping it keeps pytest as the reported cause.

This also contains a live hazard: **`coverage.xml` is committed to the repo** and is badly stale (`<source>/github/workspace</source>`, no `sql_expression.py` entry at all). If pytest ever fails before writing it, diff-cover reading that file reports "No lines with coverage information" and exits 0. `success()` keeps the gate away from that file. Untracking `coverage.xml`/`pytestresult.xml` is worth doing separately.

### What I did about `minimum_coverage: 75`

Kept the report — the per-file table is a genuinely useful annotation — but stopped it implying a gate, doing **both** available fixes rather than picking one:

- `fail_below_threshold: false` stated explicitly, so the reason it can't gate is on the page instead of hidden in the action's defaults.
- `minimum_coverage` dropped `75` → `57`, the repo's actual global coverage. The number only drives the check/cross icons, so a cross now means "below the repo average" — a real debt signal — rather than measuring against an aspiration nothing enforces.
- A comment above the step says it is report-only and points at the real gate.

Raising `fail_below_threshold: true` was rejected: that reinstates the global gate this PR argues against.

## How it was verified — by running, not reasoning

All of the following was run locally against a real `pytest` run (758 passed, 2 skipped) before pushing.

**Path resolution.** Fresh `coverage.xml` emits `<source>` = repo root with repo-relative `filename="plaidcloud/utilities/sql_expression.py"`, which lines up with git paths, so no `--src-roots` is needed. (`--src-roots` would not have helped regardless — `diff-cover --help`: *"List of source directories (only for jacoco coverage reports)"*.)

**Positive case** — the change in this PR, reproduced identically in CI:

```
-------------
Diff Coverage
Diff: origin/master...HEAD, staged and unstaged changes
-------------
plaidcloud/utilities/sql_expression.py (100%)
-------------
Total:   5 lines
Missing: 0 lines
Coverage: 100%
-------------
```

**Negative case** — appended an uncovered function, regenerated coverage:

```
Failure. Coverage is below 80%.
-------------
plaidcloud/utilities/sql_expression.py (66.7%): Missing lines 2033-2035
-------------
Total:   9 lines
Missing: 3 lines
Coverage: 66%
-------------
EXIT=1
```

The gate fails. Note the first attempt at this negative test **passed** because `coverage.xml` was stale — precisely the false-pass mode, reproduced and understood rather than assumed.

**A brand-new untested module IS caught** (checked because it's the obvious hole in any patch-coverage gate). Added `zz_probe_module.py` that nothing imports:

```
plaidcloud/utilities/sql_expression.py (100%)
plaidcloud/utilities/zz_probe_module.py (25.0%): Missing lines 2-4
-------------
Total:   9 lines
Missing: 3 lines
Coverage: 66%
EXIT=1
```

It is traced because `--doctest-modules` imports every module under rootdir and `--cov=.` sources the whole tree, so the module lands in `coverage.xml` with 0-hit lines rather than being absent. Two caveats found while testing this: the module must be **tracked** (diff-cover diffs git, so an untracked file is invisible), and it must be present for a **full** suite run (running a single test file does not collect it).

**Docs-only diff** — README-only change reports "No lines with coverage information in this diff" and exits 0, so documentation PRs are not blocked.

**Coverage granularity.** Only statement-*start* lines appear in `coverage.xml`; continuation lines of a multi-line call are absent entirely. A change touching only continuation lines would measure zero lines and silently pass, so the deliberate source change was chosen to land on covered statement lines.

## The deliberate source change

`get_update_query` built its values dict through a dict comprehension over a nested list comprehension that concatenated tuples:

```python
values = {
    col_name: value
    for col_name, include, value in [
        (tc['source'],) + get_update_value(tc, table, dtype_map, variables)
        for tc in target_columns
    ]
    if include
}
```

Replaced with the plain loop its sibling `get_update_rewrite_query` already uses, so the two functions now read identically. No behaviour change; full suite green; `ruff` findings on the file unchanged at 7 before and after.

## Notes

- Do **not** merge without a human — this changes the merge gate itself.
- `5monkeys/cobertura-action@master` is still a floating ref on a third-party action. Out of scope here, but worth pinning.


inviscid (commented):
✅ Clean Sonnet review — **APPROVE** · **not merged** (human approval required: this changes the merge gate itself)

Reviewed against the real CI run and `diff-cover`'s source, not the PR description.

**The gate is real, not vacuous.** Run 30752256391 shows `Check Patch Coverage` executing after `Perform Tests` and computing `Total: 5 lines / Missing: 0 / Coverage: 100%` against genuinely diffed lines in `sql_expression.py`. `diff_cover/diff_cover_tool.py:439` confirms `percent_covered >= fail_under`, so 80 is inclusive.

**`BASE_REF` is injection-safe** — passed via `env:` and referenced as `"${BASE_REF}"` in the shell body, never interpolated through `${{ }}`.

**The committed-and-stale `coverage.xml` is not a live false-pass path here.** Both tracked artifacts are regenerated by `pyproject.toml`'s `addopts` on every pytest run, and a pytest failure fails the job before the gate's default `success()` would let it run. It remains a footgun for anyone running diff-cover locally without a fresh pytest.

Non-blocking findings, recorded and accepted:

- **A brand-new module never imported by any test does not appear in `coverage.xml` at all**, so diff-cover reports "no lines with coverage information" and passes vacuously. Inherent to statement-coverage patch gates rather than introduced here — but it means the CHANGELOG line "new code cannot land untested" overstates the guarantee. Worth softening to something like "changed lines in traced modules must be covered".
- `coverage.xml` / `pytestresult.xml` staying tracked and stale is out of scope here but worth untracking for hygiene.
- `5monkeys/cobertura-action@master` and `andersy005/gh-action-py-liccheck@main` remain floating refs next to the newly-pinned `diff-cover==10.4.1`. Pre-existing, flagged by the author as separate work.

inviscid (commented):
🤖 Correction to my review above — one finding's **mechanism was wrong**, though its conclusion still held.

I wrote that "a brand-new module never imported by any test does not appear in `coverage.xml` at all, so diff-cover reports no lines and passes vacuously." That is **false for this repo**, and it was tested rather than argued: a `zz_probe_module.py` that nothing imports was added and the full suite run —

```
plaidcloud/utilities/sql_expression.py (100%)
plaidcloud/utilities/zz_probe_module.py (25.0%): Missing lines 2-4
Total:   9 lines / Missing: 3 lines / Coverage: 66%
EXIT=1
```

The gate **catches** it. `--doctest-modules` imports every module under rootdir and `--cov=.` sources the whole tree, so an unimported module lands in `coverage.xml` with zero-hit lines rather than being absent. Two real conditions surfaced while testing: the file must be **git-tracked** (diff-cover diffs git, so untracked files are invisible — which is what made the first attempt look like a pass), and it needs a **full** suite run, since a single test file does not collect it.

So the "new module escapes the gate" hole does not exist here. The CHANGELOG reword was still right to make — "new code cannot land untested" was overstated on the branch-protection grounds alone — but the false mechanism was kept out of it.

Recording this because a review comment is the audit trail, and an accepted finding that is wrong about *why* is worth more harm later than one that is simply wrong.

PR #359, opened 2026-08-03
fix(dialect): raise on a missing or unloadable datastore dialect instead of guessing (sc-23700)

inviscid:
## ⚠️ Deploy ordering — read first

**workflow-runner https://github.com/PlaidCloud/workflow-runner/pull/1162 (sc-23713) must deploy before or with this PR.**

The `query.py` change below converts *silently-wrong SQL* into a *loud failure* on any image with an incomplete driver set. Today `solver-runner` ships only databend + starrocks and `report-runner` ships none at all, so shipping this alone would newly break those paths. https://github.com/PlaidCloud/workflow-runner/pull/1162 adds the missing drivers to both images plus a build-time assertion that every stage can load all four. This is the ordering that turns a correctness fix into an outage if it goes first.

## Scope

Two sites in this repo, both members of the "guess a dialect rather than fail" family. The live instance of the same family — the Forecast job pod compiling Greenplum SQL on every tenant — is https://github.com/PlaidCloud/plaid/pull/6869 / https://github.com/PlaidCloud/workflow-runner/pull/1161.

### 1. `query.py` `Connection.__init__` — the worst of the family

```python
try:
    dialect_cls = registry.load(_dialect_kind)
except:
    dialect_cls = registry.load('postgresql')
```

A missing driver did not raise — it compiled **Postgres SQL against a Databend or StarRocks warehouse**. Strictly worse than the `dialect or 'greenplum'` defaults, which at least died at `registry.load` because `sqlalchemy-greenplum` is installed nowhere: this one produces plausible-looking SQL that may well parse and answer the wrong question, with nothing logged. The bare `except` also swallowed `KeyboardInterrupt`/`SystemExit`.

Now raises `RuntimeError` naming the dialect, chained `from` the underlying `NoSuchModuleError`, which is the only exception caught.

**This is the path the Solver actually uses** (`solver.py:154` → `get_dataframe_by_query`), which corrects what https://github.com/PlaidCloud/workflow-runner/pull/1161's body originally said about the Solver's failure mode: on an image lacking the driver it silently compiled Postgres rather than raising `NoSuchModuleError`. Note the Solver's dialect *source* is still wrong for a project on a non-default lakehouse — `rpc.analyze.query.dialect()` → `dst.get_dialect(None, None)` → the cluster default engine — which remains sc-23158 WS-J4.

### 2. `analyze_table.compiled` — `dialect='greenplum'` default

Raises `ValueError` when the dialect is missing or blank. Every `compiled()` call site in `plaid` already passes a real dialect and `workflow-runner` does not import this module, so no current caller is broken.

- `send_query` keeps its original four-parameter signature. An earlier revision added a `dialect` argument with no caller — the speculative-optional-parameter anti-pattern — so it is reverted; `compiled(query)` raises there, the correct outcome for a function with no dialect source. Callers should use `plaidcloud.utilities.query.Connection`.
- **Removed the deprecated `AnalyzeTable.head`**, `send_query`'s only in-repo caller. Flagged in review as a regression from the raise; it is not — it was already dead. It calls `send_query(self._rpc, self.project_id, query)`, three positional args against `send_query(project, query, params=None, rpc=None)`, so `rpc` is `None` and a fresh `Connect()` is built, `project` is the RPC object and `params` is the query. `self.project_id` is a plain method on the class (`__new__` assigns `_project_id`, never `project_id`), so it evaluates to a **bound method**, not a string — verified against the installed package (`inspect.isfunction(AnalyzeTable.project_id) is True`). It could never have returned rows.

`AnalyzeTable` itself is left in place (deprecated in favour of `plaidcloud.utilities.query.Table`); deleting a published class is an owner call.

## Tests

- `test_query.py`: `test_dialect_falls_back_to_postgresql` asserted the old contract and is **replaced** by `test_unloadable_dialect_raises_and_does_not_substitute_postgresql` — an unloadable dialect raises and names itself, and no Postgres substitution happens.
- `test_analyze_table_dialect.py`: blank and omitted dialect both raise.
- `test_sql_expression.py` gets `compiled = functools.partial(_compiled, dialect='postgresql')`. Not `'greenplum'`: pinning the engine being removed would keep a test-only dependency on the `sqlalchemy-greenplum` extra (`pyproject.toml:49`) and make a dead engine the SQL-generation reference. Postgres is a proven drop-in — 733 passed either way, no assertion changed.

With both sites converted, the `fallback-trio-other-repos` ledger entry in plaid's `test_dispatch_unknown_dialect.py` is retired in https://github.com/PlaidCloud/plaid/pull/6869 and replaced by a pointer to the assertions here.

## Companion PRs

- plaid: https://github.com/PlaidCloud/plaid/pull/6869
- workflow-runner (this ticket): https://github.com/PlaidCloud/workflow-runner/pull/1161
- workflow-runner (sc-23713, **deploy first or together**): https://github.com/PlaidCloud/workflow-runner/pull/1162

**Pin coordination:** inert until a release + pin bump. `plaid` pins 1.25.0, `workflow-runner` 1.26.0.

**Flag for whoever bumps the pin:** `plaid`'s `data_explorer/sql_generator.py:141,143,150` still carry `dialect or 'greenplum'`, so that guard becomes **load-bearing** post-bump — a `None` dialect reaching `compiled(...)` at `sql_generator.py:442` will start raising where it previously produced Greenplum SQL. Both live callers pass a real dialect today; `expression.py:61` still defaults its own parameter to `'greenplum'`. Tracked by the `sql-generator-dialect-default` ledger entry in plaid.

## Validation

`pytest plaidcloud/utilities/tests/` — 733 passed, 2 skipped (both pre-existing).


inviscid (commented):

Turns two silent-dialect-guessing bugs into explicit failures: `analyze_table.compiled`'s `'greenplum'` default (a driver installed nowhere) becomes a `ValueError`, and `query.Connection.__init__`'s bare `except: registry.load('postgresql')` — which silently compiled Postgres SQL against a Databend/StarRocks warehouse — becomes a `RuntimeError` chained from `NoSuchModuleError`. Reviewer verified the exception type empirically, confirmed `AnalyzeTable.head()` was already unreachable dead code (positional-arg mismatch bound `query` to a bound method), and confirmed the replaced `test_query.py` case tightens rather than removes coverage.

Two cosmetic nits only (a stray blank line; a narrower `except` that lets an installed-but-broken driver propagate — the desired outcome).

⚠️ **Merge ordering:** held until `PlaidCloud/workflow-runner#1162` merges, which ships the dialect drivers this raise depends on. Merging this first would convert silently-wrong SQL into failed steps on every report and solver pod.

inviscid:
Ordering constraint satisfied: `PlaidCloud/workflow-runner#1162` merged at 2026-08-03T22:20:26Z, shipping the dialect drivers into the report and solver images. Merging.

PR #361, opened 2026-08-04
feat(dialects): declare the required SQL dialect drivers so every consuming image can check itself

inviscid:
## The problem

Since sc-23700 (#359), `query.Connection.__init__` **raises** when the project's datastore dialect has no driver in the image, instead of silently substituting `postgresql` and compiling Postgres SQL against a Databend or StarRocks warehouse.

That is the right answer at the wrong moment: it fires at query time, on a customer's step, in whichever image happens to be short a driver. The mitigation so far has been a deploy-ordering rule — ship the driver images before bumping the utilities pin. Ordering is a process guarantee, and those fail.

Four independent pins consume this library (plaid `requirements.txt`, plaid `requirements-import.txt`, plaid's dead `panel_requirements.txt`, workflow-runner `requirements-base.txt`) and only workflow-runner was guarded, by its own private copy of the dialect list.

## The fix

`plaidcloud/utilities/dialects.py` declares the required set **next to the code that enforces it**, plus `assert_dialects_available(image, requirements_file)` and a CLI every consuming image runs at build time:

```
python -m plaidcloud.utilities.dialects <image> <requirements-file>
```

**This is what makes ordering irrelevant.** The list ships in the same wheel as the raise, so bumping the pin to a version needing driver X makes any image lacking X fail its own build — in whatever order the pin bump and the driver addition happened.

## Judgement calls

**Static list, not derived from what is installed.** The failure mode being guarded is *absence*; a set read off the environment defines the bug away, because whatever is installed is then "required". The list is a declaration, and its only maintenance is adding an engine — exactly when review is wanted.

**Greenplum is deliberately absent** (owner decision 2026-08-03). It is in no requirements file anywhere, and the dead default to it is where this defect family started. `sqlalchemy-greenplum` stays in the `test` extra only because ~50 existing `sqlalchemy_functions`/`sqlalchemy_views` assertions still render against it — noted, not touched here.

**Build-time only; no import-time or startup assertion.** Rejected:
- *Import-time* (in `__init__.py`) would make the library unimportable anywhere without all four drivers — this repo's own CI without the `test` extra, the Panel base image, PlaidXL, notebooks — and would add four `registry.load()` entry-point scans to every import.
- *Startup / health-check raise in the consumer* would CrashLoop a whole tenant's pod over one warehouse's driver, which is strictly worse blast radius than the per-project raise it duplicates.
- The "already-deployed old image + new pin" case it would cover **does not exist** for a baked-in library: the pin and the wheel ship inside the same image. The one place a new pin meets an older base is the Panel Kaniko build (a user's `requirements.txt` installed over `platform-panel-base`) — and that is itself a build, so the same build-time guard covers it once that base image adopts it.

## The error message is the deliverable

Someone hitting this at 3am learns which image, which dialect, which pip package, and which file to edit — with no traceback burying it:

```
rpc-import: missing 1 of 4 SQL dialect drivers.
This image cannot compile SQL for every warehouse PlaidCloud supports, so any
project on one of the engines below dies at plaidcloud.utilities.query.Connection()
— at query time, on the customer, with the image deployed green (sc-23700).

FIX: add the pip package(s) below to requirements-import.txt, then rebuild rpc-import.

  dialect 'databricks'  ->  pip package 'databricks-sqlalchemy'   [Can't load plugin: sqlalchemy.dialects:databricks]

The required set is plaidcloud.utilities.dialects.REQUIRED_DIALECTS — it travels
with the plaidcloud-utilities pin, so this is a driver gap in the image, never a
reason to relax the check.
```

## Proof the guard fires

A scratch `python:3.12-slim` build mirroring plaid's builder stage, with three of the four drivers installed, fails with exactly that message:

```
ERROR: process "/bin/sh -c python -m plaidcloud.utilities.dialects rpc-import requirements-import.txt" did not complete successfully: exit code: 1
```

The same image plus `databricks-sqlalchemy==2.0.10` builds clean and prints
`rpc: all SQL dialect drivers load (databend, databricks, snowflake, starrocks)`.

Also verified the module imports on a bare slim image with no generated locales — `plaidcloud.utilities.__init__` reaches only `udf_utility_loader` and `udf_helpers`, never `data_helpers` and its module-level `locale.setlocale`.

## Consumers (separate PRs, both blocked on this releasing as v1.28.0)

- **plaid** — one `RUN` line in the shared Dockerfile guards both `rpc` and `rpc-import`.
- **workflow-runner** — converges onto this list and deletes `workflow_runner/utils/dialects.py` plus the duplicate dialect→package map in `tests/test_image_dialects.py`, so no second definition survives.

## Runtime impact

None. Nothing on any request path calls this; `Connection.__init__` is unchanged.


inviscid (commented):

Verified end-to-end rather than by inspection. The reviewer built a clean venv, installed the real `databend-sqlalchemy` / `databricks-sqlalchemy` / `snowflake-sqlalchemy` / `starrocks` packages, and confirmed **all four `REQUIRED_DIALECTS` keys resolve via `registry.load()` to the correct dialect classes** — with a bogus-dialect positive control proving the mechanism actually discriminates rather than passing vacuously. Then it uninstalled `databricks-sqlalchemy` and ran `python -m plaidcloud.utilities.dialects rpc-import requirements-import.txt` for real: exits 1 with exactly the documented message, no traceback.

Two properties that would have made this worse than no guard, both confirmed:
- **It runs on a bare builder image** — `plaidcloud/utilities/__init__.py` imports only `udf_utility_loader` (stdlib-only) and an empty `udf_helpers/__init__.py`, so it never reaches `data_helpers`'s module-level `locale.setlocale`.
- **`SystemExit` is confined to `__main__`**; `assert_dialects_available` raises a plain `RuntimeError`, so a programmatic importer is not surprised by a process-exit exception. Clean split, and built-in exceptions per CLAUDE.md.

Greenplum is absent from the required set and covered by an explicit test — correct, since it was removed as a lakehouse engine and `sqlalchemy-greenplum` is in no requirements file anywhere; that dead fallback is the root cause of this bug family. It remains only in the `test` extra, which is out of scope.

Tests assert behaviour, not implementation: the unmocked success path doubles as a positive control, and the failure-message test checks the four load-bearing facts (image, dialect, package, requirements file) rather than snapshotting the string.

Nits, non-blocking: `REQUIRED_DIALECTS['databend']` uses the hyphenated package name while `pyproject.toml`'s `test` extra uses the underscore form (PyPI normalizes both); and `unloadable_dialects()` catches `(NoSuchModuleError, ImportError)` — one type wider than `query.py`'s precedent, plausibly deliberate for partially-broken installs, worth a one-line comment.

PR #362, opened 2026-08-04
fix(sql_expression): resolve a whole target-column list's dtypes once (sc-23870)

simozzy:
**Library half of sc-23870.** Needs the matching `workflow-runner` change ([workflow-runner#1173](https://github.com/PlaidCloud/workflow-runner/pull/1173)) and a release + two pin bumps before it reaches a customer. Sequencing at the bottom.

## The defect this closes

sc-23460 added `_target_dtype`, called by `get_from_clause` while building the CAST. That fixed the **emitted SQL** and nothing else.

A step also *declares* the table it writes, from the same config, and that path reads the raw dtype:

```python
# get_table_rep, before this PR
sqlalchemy.Column(sc[column_key], sqlalchemy_from_dtype(sc['dtype']))
```

`sqlalchemy_from_dtype` stringifies before its regex lookup, so `None` becomes the key `'none'` and raises. In `workflow_runner`'s `frame_lookup` that call comes **before** `analyze.table.touch`, so an untyped Lookup/Join died there — one function earlier than the half sc-23460 fixed, and the reason that ticket's fix never made an already-saved step run.

## What this adds

`resolve_target_dtypes(target_columns, source_column_configs, tables=None, table_numbering_start=1, *, tables_by_alias=None)` — the list form of the same rule, so a caller resolves once and hands one answer to every consumer.

Plus a `text` backstop in `get_table_rep` for a **null** dtype. An **absent** `dtype` key still raises, as it always has: that is a malformed config rather than the untyped-column shape a step form can write, and most of this function's ~86 call sites are representing a *source* table, where quietly declaring text would mistype a column the query reads.

## Read this before using it: resolving freezes the answer

`_target_dtype` returns any truthy dtype unchanged. So once `resolve_target_dtypes` has written a value in, **a later `get_from_clause` — even one holding the real table objects — will not re-derive it.**

That matters because the caller decides the fidelity:

| resolved with | ambiguous name across sources | 
| --- | --- |
| `tables` passed | the dtype of the table the query actually reads |
| no `tables` | `text` |

The runner half resolves at `TransformHandler.__init__`, where no tables exist yet, so it takes the second row. **That is a semantic narrowing, not a wider type** — a numeric column declared and cast as text loses arithmetic and ordering. It is still the right call against the two alternatives:

- **Guessing a side** yields `CAST(<text column> AS NUMERIC)`, which a warehouse can reject. The reverse cannot fail.
- **Leaving the null in place** puts the declaration and the CAST back into disagreement — the exact bug being fixed.

It is not a regression: a config that hits this crashes on master today, so there is no non-crashing baseline to regress from. But it is a trade, and an earlier revision of this PR described it as "never wrong, sometimes wider than it needs to be", which was wrong on both counts. `test_resolving_without_tables_freezes_the_lower_fidelity_answer` pins it so it stays visible.

Returns **shallow** copies, so a config that is round-tripped or persisted keeps its dtype — typing a column permanently stays the user's decision, via the step form. Nested values are shared with the originals; nothing here mutates them.

## Adversarial review

Two rounds by a `sonnet` subagent, run to convergence.

**Round 1 — one blocker, and it was a design fault rather than wording.** The freeze above was undocumented, and the docstring asserted the declaration and the CAST "cannot disagree". Also found the backstop swallowing an absent key, not just a null. Both fixed in `111f91a`; the review's suggested assertion fix (`assertIsInstance(..., sqlalchemy.Numeric)`) was itself wrong — `PlaidNumeric` is a `TypeDecorator`, not a `Numeric` subclass — and was caught by running it.

**Round 2 — no blockers**, and it confirmed the backstop's `KeyError` on an absent key is a *restoration* of master's bare subscript rather than new breakage, via a ~150 call-site audit across all three repos. Remaining polish in `beae19a`: deleted a near-circular test whose comment claimed it checked something it never built, restated the freeze test through the public API instead of poking `_target_dtype`, and added the one pass-through (`table_numbering_start`) that had no coverage.

Round 2 also flagged two *pre-existing* bare `t['dtype']` subscripts in `plaid`'s `import_tools.py` (lines 267, 326), reachable via REST/JSON-RPC/MCP. Same family, out of scope here, not yet ticketed.

## Verification

- `pytest plaidcloud/utilities/tests/` — **771 passed, 2 skipped**.
- `ruff check --select E9,F63,F7,F82` clean.
- Every new line is exercised, which is what this repo's real gate (`diff-cover --fail-under=80` patch coverage) measures. `pylint --errors-only` gains two `E1131` union-annotation false positives — the same one already firing 32 times in this file, including on `_target_dtype`'s own signature immediately above.

## Sequencing

1. **This PR** merges and gets a release.
2. `plaid`'s `requirements.txt` **and** `workflow-runner`'s `requirements-base.txt` both bump to it. Two pins govern this, not one — bumping only plaid's leaves every step *run* on the old library, which is the trap sc-23460 hit.
3. The runner PR's tests **skip** until that bump lands (they check for `resolve_target_dtypes` and skip with a message naming why). Watch for skips-turning-to-passes rather than reading the badge.


plaidcloud/utilities/sql_expression.py
```diff
@@ -813,6 +813,54 @@ def _target_dtype(
...
+    arithmetic and ordering. It is chosen over guessing a side, which yields
+    ``CAST(<text column> AS NUMERIC)`` — rejectable by the warehouse where the
+    reverse is not — and over leaving the null in place, which puts the
+    declaration and the CAST back into disagreement.
+
+    Shallow copies, so the caller's configs keep their dtype: a config is
+    round-tripped and may be persisted, and a resolved dtype is a run-time
+    reading of an untyped column, not a decision to type it permanently —
+    that stays the user's, via the step form. Nested values are shared with
+    the originals; nothing here mutates them.
```
rad-pat: I was hoping to read the comment and gain some insight as to what the method did, but reading this just confuses me. It mentions that get_from_clause uses it, yet I can't see that in the PR?

plaidcloud/utilities/sql_expression.py
```diff
@@ -1034,7 +1082,20 @@ def get_table_rep(table_id: str, columns: list[dict], schema: str, metadata: sql
...
+                # dtype where this cannot (it is given no source configs). A
+                # null dtype that still arrives here would otherwise raise
+                # `RegexMapKeyError: 'none'` — sc-23870. Inert once resolved:
+                # a resolved dtype is truthy.
+                #
+                # Subscript, not `.get`: an *absent* dtype key is a malformed
+                # config rather than the untyped-column shape a step form can
+                # write, and most of the ~86 call sites are building a rep of a
+                # *source* table, where quietly declaring text would mistype a
+                # column the query reads. That keeps raising, as it always has.
```
rad-pat: I thought commenting was supposed to be concise. More comment than code is not cool. I blame Claude. I'm sure this can be a one-liner comment.

rad-pat (changes requested):
Large comments and confusion from me.

plaidcloud/utilities/sql_expression.py
simozzy: Fair, and thanks for pushing on it — the confusion was the docstring's fault twice over.

**Your actual question first:** `get_from_clause` does *not* call `resolve_target_dtypes`. It calls `_target_dtype`, and has done since sc-23460 — which is why you couldn't find it in this diff. The sentence said "`_target_dtype` resolves one column, and `get_from_clause` calls it", and "it" was doing far too much work. Nothing was missing from the PR; the wording invented a call. It's gone.

**Second problem:** three of the four paragraphs argued a trade-off before ever saying what you'd call the function for. Now 28 lines → 12, which is mid-pack for this file (median is 8 across its 38 docstrings; this was the second-longest of the lot, sitting next door to `get_from_clause`'s one-liner). What survives is what changes how you call it — resolve once and hand the result to both consumers, and pass `tables` if you have them because resolving freezes the answer.

The `CAST(<text column> AS NUMERIC)` argument didn't need restating here at all — it's already in `_target_dtype`'s docstring, where the rule actually lives. The freeze cost stays pinned by `test_resolving_without_tables_freezes_the_lower_fidelity_answer`, and the long version is in the PR description if anyone needs it.

plaidcloud/utilities/sql_expression.py
simozzy: Agreed — 13 lines of comment on a one-character change is indefensible. One line now:

```python
# Null dtype → text, so the declaration agrees with the CAST (sc-23870); an absent key still raises.
sqlalchemy_from_dtype(sc['dtype'] or 'text'),
```

One thing worth saying about what I cut, since it's the only bit that genuinely leaves the source file. The deleted block was justifying a deliberate asymmetry — a **null** dtype falls back to text, an **absent** key still raises — and the justification was a call-site audit (~86 of them; most are building a rep of a *source* table, where quietly declaring text would mistype a column the query reads). The one-liner names the asymmetry, `test_an_absent_dtype_key_still_raises` pins it, and the audit stays in the PR description. If you'd rather that reasoning lived in the code after all, say so and I'll put a short version back — but it read to me as PR-review context rather than something the next reader of this line needs.

Same treatment applied to the test comments in this PR, since they had the same problem and the test names were already saying it.

rad-pat (approved):
Lovely jubbly

PR #364, opened 2026-08-07
Move the license gate to the shared Trivy action

rad-pat:
Replaces `liccheck` with the shared
[`license-check`](https://github.com/PlaidCloud/plaidcloud-github-actions) action.

Story: https://app.shortcut.com/plaidcloud/story/24172

## No setuptools pin here

Unlike plaid, plaid-rpc, plaidcloud-cp-rest, workflow-runner and RediStomp, this
repo never pinned `setuptools` to keep liccheck alive — `requirements.txt` has no
entry and `pyproject.toml` uses unpinned `["setuptools", "setuptools-scm"]`. So
there is nothing to unblock here, only a dead tool to replace: liccheck imports
`pkg_resources`, which setuptools 82 removed, and its last release was September
2023.

Worth noting for the wider effort: `plaidcloud-rpc>=1.11.0` here is a **floor**,
so it already accepts the 1.15.0 that carries the unpin. This repo was never a
blocker in the `plaid → plaid-utilities → plaid-rpc` chain.

## Two policy entries not carried over

This repo's `liccheck.ini` authorised the literal string **`unknown`** (and
`mit and psf-2.0 unknown`). That approves any package whose licence liccheck
could not determine — the whole of `level: cautious` disabled by one line, and
the reason a shared policy is worth having.

**I verified it was covering nothing.** Built this repo's real dependency tree
on Linux and ran the shipped gate with no waivers:

```
Policy allows 73 license spellings and 0 package waiver(s) from 1 file(s). Trivy reported 41 packages.
All dependency licenses are approved.

   packages with NO license: none
```

Every package resolves to an approved licence, including the compound forms
`BSD-3-Clause AND 0BSD AND MIT AND Zlib AND CC0-1.0`,
`MPL-2.0 AND (Apache-2.0 OR MIT)` and `MIT AND PSF-2.0`, which clear on their
atoms rather than needing entries of their own.

So this migration needs **no repo-local waivers**, and the stricter policy costs
nothing here.

## Scope and shape

The check gets its own job rather than a step on the lint job — that one installs
`.[test]` plus `ruff` and `diff-cover`, and the gate should police what ships, not
the tooling. No extras are passed: the base dependencies are what plaid and
workflow-runner install.

`fetch-depth: 0` because `setuptools_scm` derives the version from tags, and a
depth-1 clone has none — the install would fail before anything was scanned.

The gate is stricter than liccheck was in two other ways worth knowing: an
*undeclared* licence fails, and a dual licence needs **both** halves approved, so
a package we could legally take under the permitted half stops for a person
rather than being chosen for silently.


simozzy (commented):
Reviewed the diff, the shared action at the revision `@v1` resolves to today, the
policy file, this PR's own CI log, and how the four consuming repos install this
library. Good change, well-argued PR body — the substantive claims hold up when
checked. Two things to fix before merge, both cheap.

---

## 1. The green run is stale — `@v1` moved ten minutes after it

`Check Dependency Licenses` passed at `14:10:28Z` (41 packages, 0 waivers, all
approved — matches the PR body exactly). The `v1` tag was re-pointed at
`14:20:07Z`, to `ffa29f1b` *Add license assertions, and approve CNRI-Python*.
The run therefore exercised `6d5eba8c`, not what `@v1` resolves to now.

The delta looks additive (assertions + one extra approved spelling), so I doubt
it breaks — but "the gate is green" is currently a statement about a revision
that is no longer what runs. Push an empty commit or re-run the job.

The general form is worth a decision at org level rather than per repo: a
compliance gate pinned to a moving major tag can change what it enforces with no
PR in the repo it enforces against. It just did. `plaid-rpc` has the same shape.
Not a blocker for this PR — pinning to a SHA is the org's call, not yours here —
but the re-run is.

## 2. Dead liccheck config left behind in `pyproject.toml`

The PR deletes `liccheck.ini` and the changelog says the tool is gone, but
`pyproject.toml` still carries ~55 commented lines of the same policy under
`# Can be moved from liccheck.ini if we want`, including:

    #    "gnu general public license (gpl)",

which is exactly what the org policy deliberately refused to carry over. A dead
tool's config surviving as a comment is how it comes back — someone uncomments
the block in six months and re-approves GPL and the `unknown` family in one
edit. Delete it in this PR; it is the same change. (`plaid-rpc`'s pyproject has
the identical leftover — worth the same one-liner there.)

## 3. The `fetch-depth: 0` comment claims more than is true

    # setuptools_scm derives the version from tags, and a depth-1 clone
    # has none — the install would fail before anything is scanned.

`plaid-rpc` uses `setuptools_scm` the same way, calls the same action, checks out
with a bare `actions/checkout@v7` and no `fetch-depth` — and its license job is
green. So the install does not fail without tags; you get a fallback
`0.1.devN+g<sha>` version instead of this run's `1.30.1.dev3+g3abf1396`.

Keep the `fetch-depth: 0` — it is nearly free here and a real version is better
than a fabricated one — but reword to what actually happens. Alternatively, if
you have evidence the install genuinely dies, then `plaid-rpc`'s job is passing
on a broken premise and wants the same two lines. One of those is true and it is
worth knowing which.

## 4. Nothing actually enforces either gate (not a regression — flagging it)

`master` has no branch protection (`/branches/master/protection` → 404) and no
rulesets (`/rulesets` → `[]`). The old check failed a step inside the lint job;
the new one is its own job. Both are advisory either way, so this PR changes
nothing — but if the intent of sc-24172 is a *gate*, the required-check config is
the other half of it, and moving the check into its own job is the moment to add
`Check Dependency Licenses` to the required list. Worth confirming the 404 isn't
just my token's permissions.

---

## Claims I checked and can confirm

Recording these so the next reader doesn't redo them:

- **"No extras" is right.** `plaid/requirements.txt:85`,
  `plaid/requirements-import.txt:70` and `workflow-runner/requirements-base.txt:28`
  all take a bare `plaidcloud-utilities==1.30.0`; an org-wide code search for
  `plaidcloud-utilities[` returns no consumer at all. Base deps are what ships.
- **The `test`-extra dialect drivers aren't lost.** `starrocks`,
  `databend-sqlalchemy`, `databricks-*`, `snowflake-sqlalchemy`, plus `fastavro`
  and `yxdb`, are all direct entries in plaid's own `requirements.txt`, so their
  licences are scanned by plaid's gate rather than by nobody.
- **The dropped package waivers were dead.** `hdbcli` — the one genuinely
  proprietary entry in the deleted `[Authorized Packages]` — is not in plaid's
  requirements at all.
- **No scan-set narrowing.** `pyproject.toml` takes `dependencies` from
  `requirements.txt`, so `uv pip install .` covers exactly what liccheck was
  pointed at, plus transitives.
- **The real migration risk was LGPL, and it's covered.** `chardet` is LGPL-2.1
  and the org policy carries the whole family explicitly.

## Nit

The PR body quotes "Trivy reported 41 packages" as coverage evidence; uv resolved
28. The extra 13 are vendored `dist-info` nested inside other packages, which the
action's own completeness check deliberately excludes. The number is a faithful
quote of the tool, it just isn't the dependency count if a reader takes it as one.

---

**Verdict:** approve once the license job has been re-run against the current
`@v1` and the commented liccheck block is out of `pyproject.toml`. Findings 3
and 4 are wording and process, not code.

simozzy (dismissed):
Claude asked for a couple of changes.

rad-pat:
All four fixed or answered — thanks, findings 1 and 2 were both real misses.

**1. Stale green run — fixed, and the cause was worse than "it moved".** You were
right that the quoted run exercised `6d5eba8c`, not current `@v1`. The reason no
re-run had happened: this workflow declares no `types:`, so it gets the
`pull_request` defaults — `opened`, `synchronize`, `reopened`. **`ready_for_review`
is not among them**, so marking the PR ready never re-triggered it. Pushing the
fixes did. Current run is green against v1.1.0, and the wording confirms it:

```
Policy allows 74 license spellings, 0 waiver(s) and 0 assertion(s) from 1 file(s). Trivy reported 41 packages.
```

(v1.0.0 printed `73 license spellings and 0 package waiver(s)` — no assertions
existed yet.)

On the general point: agreed, and it is a real property rather than an oversight.
`@v1` is deliberately moving so a policy change reaches every repo without ten
PRs — that was the whole argument for a shared policy. The cost is exactly what
you describe. It is written down in the action's README, but it deserves an org
decision rather than a README line, so I have not changed it here.

**2. Commented liccheck block — deleted.** 64 lines, and you are right about the
failure mode: uncommenting it re-approves `gnu general public license (gpl)` and
the `unknown` family in one edit. That block was more dangerous than the file I
did delete, since `liccheck.ini` at least announced itself. plaid-rpc has the
identical leftover — it is merged, so it needs a follow-up PR rather than a
change here, and I will raise one.

**3. `fetch-depth: 0` comment — reworded.** You are right and I checked it your
way: plaid-rpc runs `setuptools_scm` through the same action with a bare
checkout and is green, so the install does not fail. It fabricates
`0.1.devN+g<sha>`. Kept the fetch because a real version in the scan is worth it,
but the comment now says what actually happens rather than inventing a failure.

**4. Nothing enforces either gate.** Confirmed, and out of scope here — it was
advisory before this PR and is advisory after. Worth its own conversation about
branch protection across the org; I have not touched it.

**Nit — "41 packages".** Fair. uv resolved 28 distributions; Trivy reports 41
because it also reads `dist-info` vendored inside other packages. The action's
own completeness check deliberately excludes those, so 41 is a faithful quote of
the tool and not the dependency count. Reworded in the PR body to say both
numbers.

Commits squashed to one.

inviscid (approved):

Independent reproduction confirms the change: cloned head, `pip install .` in a clean venv, ran the pinned Trivy scanner + real `check_licenses.py` against org policy — 41 packages, all approved, matching the PR body. `pyproject.toml` dynamic deps from `requirements.txt` means the new action's default `package: '.'` install scans the same base dependency set the old liccheck config did — no drift. New allowlist policy is strictly tighter than the old `liccheck.ini` (drops the `unknown` whitelist; requires both halves of dual licenses). Dropped `[Authorized Packages]` entries are all dead in this repo's base tree. New `license` job correctly gates PRs to master with justified `fetch-depth: 0`.

Non-blocking:
- `.github/workflows/pr_lint.yaml` — confirm branch-protection required-status-checks on `master` are updated to require the new `license` job name; the old `gh-action-py-liccheck` check name disappears, so the gate could silently stop blocking merges if protection still lists the old name.

inviscid:
_Reviewed by Paul's automated review pass, not read by hand. Reply here if you want a human read before merging._

simozzer (approved):

PR #365, opened 2026-08-07
feat(query): Connection takes the project's dialect from its caller (sc-23158 WS-J4, 2 of 3)

inviscid:
## What

`query.Connection` takes the project's SQL dialect from its caller, and names the project when
it has to ask for one.

```python
Connection(project=..., rpc=..., dialect=config.get('datastore_dialect'))
```

`Connection.dialect` compiles and executes the SQL of roughly twenty workflow-runner transforms.
It was built from `self.rpc.analyze.query.dialect()` — **no project named**, so the server
answers for the tenant's own lakehouse. Meanwhile every workflow step is already handed a
correct per-project dialect on its run payload (`config['datastore_dialect']`). Two sources,
agreeing today because a tenant has one lakehouse, disagreeing the moment it has two, with no
rule about which a given code path reads.

The passed value now wins, and the fallback names `project_id`, so both paths answer for the
project rather than for the process. This is deliberately *not* a third source: it makes the
existing wire value authoritative and stops the derivation being tenant-global.

## No-op today

Same dialect name in ⇒ byte-identical SQL out. `test_query_dialect_parity.py` compiles the
statement shapes the consuming transforms emit — the filtered/aggregated/`HAVING`/`LIMIT`
select `frame_melt`, `frame_pivot` and `rule_assignment` build, the `INSERT … FROM SELECT` the
allocations write through, a join, a `UNION ALL`, and a `CASE`/`IN` projection — through a
`Connection` built **both** ways, and asserts the compiled SQL *and* the bound params are
equal. Parameterised over every engine in `REQUIRED_DIALECTS` this interpreter can load, plus
`postgresql`, which SQLAlchemy always registers, so the suite is never vacuous; locally that is
databend, databricks, snowflake, starrocks and postgresql — 25 subtests. A guard test asserts
`postgresql` is in the set, so a driver-less environment cannot make it pass by skipping
everything.

## Tests

Local (`/Users/inviscid/Projects/plaid/.venv`, which lacks the `greenplum` driver, so the
same 58 pre-existing failures appear on both sides):

| | master | this branch |
|---|---|---|
| passed | 713 | 718 |
| subtests passed | 447 | 472 |
| failed | 80 (58 unique ids) | 80 (58 unique ids) |

`diff` of the sorted `FAILED` lines: **identical failure sets**.

Three mutations, each caught and reverted:

| Mutation | Caught by |
|---|---|
| passed dialect ignored, always `postgresql` | `test_query_dialect_parity` (SQL differs on all five dialects) |
| fallback drops `project_id=` | `test_derived_dialect_names_the_project` |
| passed dialect ignored entirely | `test_passed_dialect_wins_and_nothing_is_derived`, `test_a_passed_dialect_that_cannot_load_still_raises` |

Null control: the unmutated run passes clean (103 passed, 25 subtests) before any mutation.

`ruff check` on the changed files reports only the two findings that are already on master
(`I001` import sorting, one `W291` that moved line numbers).

## Merge order — this is second of three

sc-23158 WS-J4 spans plaid + plaidcloud-utilities + a workflow-runner pin bump.

1. **plaid** — optional `project_id` on `analyze.query.dialect` (RPC, REST, MCP).
   **Must merge and deploy before this PR is released.** `rpc_common.rpc_method`'s wrapper
   forwards `**kwargs` straight to the function, so `dialect(project_id=…)` against a plaid
   that predates it is a `TypeError`, not an ignored extra.
2. **this PR**, released as **v1.32.0**.
3. **workflow-runner** — pin bump to 1.32.0 and pass the step's own dialect at
   `TransformHandler.__init__`. Needs this release on PyPI.

Nothing is broken at any intermediate merge: an old runner constructing `Connection` without
`dialect` still works (it takes the fallback), and a plaid that already accepts `project_id`
still answers an unqualified `dialect()` exactly as before.

## Note

`Connection` gains a **third keyword parameter**; every call site in this repo, in plaid and in
workflow-runner passes `project` and `rpc` by keyword, so no positional caller is affected.

sc-23158

---

**The three PRs, in merge order:**

1. plaid https://github.com/PlaidCloud/plaid/pull/7017
2. plaidcloud-utilities https://github.com/PlaidCloud/plaid-utilities/pull/365 (release v1.32.0)
3. workflow-runner https://github.com/PlaidCloud/workflow-runner/pull/1188


inviscid (commented):

The core change (`Connection(project, rpc, dialect=None)`, passed value wins, fallback names `project_id`) is correct and backward-compatible — a grep across plaid, plaid-utilities, plaidlink and the `udfs` repo turns up **zero positional `Connection(...)` callers**; every call site uses keywords.

**The sequencing claim was verified from source, not assumed.** `plaid-rpc:plaidcloud/rpc/remote/rpc_common.py`'s `rpc_method` wrapper does `return await function(**processed_kwargs)` with no kwarg filtering, so a server that does not declare `project_id` raises `TypeError` on an unexpected keyword. That is why the merge order was inverted, and it is accurate. Part 1 (plaid #7017) is merged as `eaf0f453d0`, and it normalizes `project_id = project_id or None` before branching — which defuses the empty-string-into-a-Redis-key crash on the server side.

**Follow-up from the first round, now closed:** the fallback's projectless behaviour was untested from the client side. Added two cases — a projectless `Connection` falls back to the rpc's own project, and when the rpc has none either the client sends `project_id=None` rather than omitting it or sending `''`. Dropping `project_id` from the fallback fails all three dialect tests. (My first draft of these asserted `None` for the projectless case, which is wrong — the code falls back to `self.rpc.project_id`. Corrected to pin real behaviour rather than my assumption.)

CI at this head: `Lint PlaidCloud Utilities w ReviewDog` **success** — this repo runs `pytest` inside that job. Verified from the full log: **811 passed, 2 skipped, 494 subtests**, with both new cases PASSED and zero skipped, and `test_query_dialect_parity.py` exercising compiled SQL and bound params across dialects behind a guard that prevents a driver-less environment passing by skipping.

⚠️ **Release precondition, not a merge blocker.** `release.yml` publishes only on an explicit GitHub `release: published` event, so merging this does not expose anything. Cutting **v1.32.0** does: every live UDF builds a bare `Connection(project=...)` with no dialect and hits this fallback, so any plaid instance that has not **deployed** #7017 will `TypeError`. Merged is not deployed — confirm rollout before publishing.

PR #369, opened 2026-08-11
fix: register the patched StarRocks dialect where Connection can see it

rad-pat:
Registers a patched StarRocks dialect so that `query.Connection` — and therefore customer UDF
code — cannot silently compile away a `CAST(… AS FLOAT)`.

**+153/−2 across 5 files.** This PR was rewritten; see the note at the bottom for what it used to
be and why almost all of it went.

## The defect

The vendored `starrocks` dialect subclasses SQLAlchemy's MySQL one, which gates float casts on
`_support_float_cast` — a property derived from `server_version_info`, populated **only by
connecting**. Compiling with an unconnected dialect reads `False`, and `MySQLCompiler.visit_cast`
**removes the cast from the statement**. Verified on this branch:

```
vendor  StarRocksDialect       →  SELECT t.a AS a FROM t          # cast gone, SAWarning only
patched PlaidStarRocksDialect  →  SELECT CAST(t.a AS FLOAT) AS a FROM t
```

A dropped cast is a wrong number on a green step, not an error.

workflow-runner has carried a patched subclass for this, but registers it as a **side effect of
importing its transform handler**. `query.Connection` builds its dialect from the same registry,
so it got the capability only when something else had already imported that module — and a
`Connection` built without it compiled the dropped cast. Since `Connection` is the surface
customer UDF code is written against, that is reachable from a UDF.

## The fix

`plaidcloud.utilities.starrocks_dialect.PlaidStarRocksDialect` lives here now and is registered
from `dialects`, which the package root imports — so **any** use of this library has it and no
consumer has to remember to import anything first. Registration stays lazy (SQLAlchemy imports
the target only when the dialect is asked for), so an image carrying no `starrocks` package still
imports these modules.

`unloadable_dialects` also catches `AttributeError` now: a name we register resolves through
*our* module, so a bad module path or renamed class raises from inside `registry.load` rather than
as `NoSuchModuleError` — uncaught, that would put a traceback in a Docker build log instead of the
message this gate exists to print. A genuinely absent vendor package still reports the same way,
because the subclass imports it.

## Tests

Five, pinning the fix, **the defect it prevents** (so dropping the registration goes red rather
than falling back quietly), the build gate's `AttributeError` path, and — in a **fresh
interpreter**, because in-process proves nothing — that importing `query` alone is enough.

That last test earned its place immediately: writing it caught the trim having gone one step too
far. With `query` no longer importing `dialects`, nothing armed the registration at all, and the
probe returned the vendor class's inherited `<property object>` instead of `True`.

Suite: **786 passing**. The 2 `test_dialects.py` failures are pre-existing and environmental —
`databricks-sqlalchemy` isn't installed locally (`starrocks` loads); CI installs it.

## Follow-up

workflow-runner should drop its own `registry.register('starrocks', …)` when it takes this
release. It runs after the runner's import of `query` and therefore wins, so until it goes the
runner keeps its local subclass. Harmless today — both classes set the same attribute — but it
means an edit to the shared one wouldn't reach the runner.

## What this PR used to be, and why it isn't

It originally moved ten modules here (`column_propagation`, `conditions`, `dynamic_selection`,
`unconverted`, `melt`, `spatial_sql`, `dimension_reference`, `errors`, plus `get_dialect` /
`compile_query` / `resolve_target_columns`) so plaid and workflow-runner could **share** them
while step execution migrates between the repos.

That was the wrong shape. The migration is a **lift-and-shift**: when an operation lands in plaid
it is *deleted* from the runner. Sharing buys nothing for code whose second copy is going away by
design — it buys a permanent library surface, a release and two pin bumps, to serve a window that
closes on its own. Those modules are duplicated inside plaid for the duration and deleted from the
runner as each operation moves.

The dialect is the one item that doesn't fit that reasoning, and it was never really a sharing
case: `query.Connection` lives in *this* library and needs the capability on its own account.

https://claude.ai/code/session_01VxgnkaCzefugmvQPVU2wiX


rad-pat:
## Adversarial review round 1 — 3 major, 4 minor, all fixed

Ran a sonnet adversarial review over the full diff, cross-checked every moved module against its workflow-runner original, and mutation-tested the load-bearing claims. Fixed everything above nit level in `4a1e161`.

### Major

**`unconverted` only *looked* shared.** Two of its three functions take a *handler* and call `handler.get_frame()`, `handler.get_dialect()` and `handler._send_query()` synchronously — none of which plaid's `StepContext` offers in that shape (its equivalents are `async` and named differently). Calling them from plaid would have raised `AttributeError` or silently passed an unawaited coroutine through synchronous logic. **Only `unconverted_step_message` moves now.** When the other two follow they should take injectable callables, the way `spatial_sql.require_wkt_geometry` and `dimension_reference.resolve_dimension_reference` already do — that changes the runner's call sites, so it isn't part of an additive move.

**Three modules claimed their tests moved and had none.** `melt`, `spatial_sql` and `unconverted` now carry theirs, each trimmed to the half whose code came. 870 → **888 passing**.

**The build-time driver gate could report a broken registration as a traceback.** Because a name we register resolves through *our* module, a bad module path or renamed class raises `AttributeError` from inside `registry.load` rather than `NoSuchModuleError` — which is exactly the Docker-build-log traceback `dialects.py` exists to replace with a message. Caught now. (The "vendor package absent" case was always fine; our subclass imports it.)

### Minor

- `errors.py` stated as settled fact that workflow-runner re-exports `UserError`. It doesn't yet — reworded to a requirement, with the live consequence named.
- The newline-flattening test was **vacuous**: it used a built `select()`, which SQLAlchemy already emits with a trailing space at every clause boundary, so it couldn't tell `''` from `' '`. Mutating `compile_query` to delete newlines left it green. It now uses raw multi-line `text()` — both the real path (a hand-authored `frame_extract_sql` step) and the case that distinguishes them. **Verified it now fails under that mutation.**
- The registration-arming test imported `query` in its body while `dialects` was already imported at module scope, so it proved nothing. Runs a fresh interpreter now.
- CHANGELOG corrected on three overstatements: which tests moved; that `get_dialect`'s falsy-dialect refusal is **new** behaviour rather than preserved (the one place this release is not strictly additive — `None` previously died on an uncaught `TypeError`, `''` was reported as a missing *driver*); and that workflow-runner's own `registry.register` still wins until removed there.

### Cleared

Move fidelity is **byte-identical** for every moved module against its runner original (`conditions`' extraction included — nothing truncated, imports correctly pruned). No circular imports. The `caplog` stand-in is **not** vacuous: removing the production warning failed exactly the 6 tests that assert it, everything else stayed green.

### Added to the follow-up list

workflow-runner must drop its own `registry.register('starrocks', ...)` when it takes this release — it runs after the runner's import of `query` and therefore wins, so until then the runner keeps its local subclass and an edit to the shared one won't reach it. Harmless today (both set the same attribute), but "one implementation" isn't yet true for the runner.

### Not fixed (nits)

An asymmetric section banner in `conditions.py`, and one test close to asserting a Python language basic.

